PR #81, opened 2018-06-03
Managing releases

### LINKED ISSUE #72: Managing releases
During an application release, all stories are initialized (i.e., started at line 1). This enables the story to listen to pubsub, register with the gateway, and queue cron jobs.

## Alpha
- Stories should unsubscribe and stop immediately.
- Stored variables are deleted (see #71).
- The Engine must accept a function which will run a specific function within that storyscript file.

```diff
 POST /story/run
 {
   'story_name': '/some/story',
+  'function': 'SomeFunction',
   'context': {...}
 }
```

# Beta
- If the story source code is unchanged in a release there may be no need to take action on that specific story.
- If the story source code changed there may be some migration policy based on the changes and services

judepereira:
Closes #72


asyncy/App.py
```diff
@@ -38,3 +42,25 @@ def apply(self):
         self.environment = self.load_file('environment.json')
         self.stories = self.load_file('stories.json')
         self.services = self.load_file('services.json')
+
+    async def bootstrap(self):
```
stevepeak: Having the bootstrap here is ok. But I did add it to the release cycle: https://github.com/asyncy/platform-bootstrap/blob/master/stories/app/release/alpha.story#L80-L83

Having this in the Engine is fine. At this time I cannot think of any reasons why not.

judepereira: @stevepeak Sure, we can remove that later on, if required. This helps for the time being and helps me iterate fast during the development process.

PR #83, opened 2018-06-11
Support for calling functions

judepereira:
Changes:
1. Removed next_line (the tree will always have ‘next’)
2. Removed start and child_block (these methods modified the tree, and no tree modifications should be permitted, given that it can remove the function definition altogether)
3. Since were using asyncio, we can close the connection inside Service itself (if not closed already)
4. Removed Handler (not useful anymore)

The alternative to using child_block is to use execute_block. 

As of right now, the integration tests are failing as the engine doesn’t have the latest asyncy/storyscript compiler. I’ll fix them shortly.


asyncy/processing/Story.py
```diff
@@ -19,41 +18,94 @@ def save(logger, story, start):
...
+
+        To execute a function completely, see Story#execute_function.
+
+        :return: Returns the next line number to be executed
+        (return value from Lexicon), or None if there is none.
+        """
+        line = story.line(line_number)
+        story.start_line(line_number)
+
+        method = line['method']
```
stevepeak: Just curios if doing a attribute lookup would be quicker and cleaner:

```py
if hasattr(Lexicon, f'async_{method}'):
  return await getattr(Lexicon, f'async_{method}')(logger, story, line)

elif hasattr(Lexicon, f'sync_{method}'):
  return getattr(Lexicon, f'sync_{method}')(logger, story, line)

else:
  raise NotImplementedError(f'Unknown method to execute: {method}')
```
> Note the methods in Lexicon must be named `sync_if`, `async_run`, `sync_function` etc.

judepereira: Not quite sure if the JIT compiler would be able to optimise that (method call vs static linking). Let's leave it at what it is right now.

I like the idea of having sync_xx and async_xx. Will implement that and merge.

judepereira: Making everything async, as it'll be required later anyway.

PR #99, opened 2018-06-24
Sandbox all file operations

judepereira:


judepereira:
Will create a new one.

PR #106, opened 2018-06-29
Mutations

judepereira:


asyncy/processing/Mutations.py
```diff
@@ -0,0 +1,135 @@
...
+        elif operator == 'shift':
+            if operand == 'left':
+                return cls.list_shift(value, -1)
+            elif operand == 'right':
+                return cls.list_shift(value, 1)
+        elif operator == 'sort':
+            value.sort()
+            return
+        elif operator == 'apply':
+            # TODO
```
judepereira: @stevepeak What is [1, 2, 3] apply supposed to do?

asyncy/processing/Mutations.py
```diff
@@ -0,0 +1,135 @@
...
+            pass
+        elif operator == 'min':
+            return min(value)
+        elif operator == 'max':
+            return max(value)
+        elif operator == 'sum':
+            return sum(value)
+        elif operator == 'reduce':
+            # TODO
```
judepereira: @stevepeak What is [1, 2, 3] reduce supposed to do?

asyncy/processing/Mutations.py
```diff
@@ -0,0 +1,135 @@
...
+        operand = story.argument_by_name(mutation, operator)
+        if operator == 'length':
+            return len(value)
+        elif operator == 'keys':
+            return value.keys()
+        elif operator == 'values':
+            return value.values()
+        elif operator == 'items':
+            value.items()
+            # TODO
```
judepereira: @stevepeak What is {a:b, c:d} items supposed to do?

asyncy/processing/Mutations.py
```diff
@@ -0,0 +1,135 @@
...
+            return value.replace(operand, new_val)
+        elif operator == 'split':
+            return value.split(operand)
+        elif operator == 'uppercase':
+            return value.upper()
+        elif operator == 'lowercase':
+            return value.lower()
+
+    @classmethod
+    def _mutate_list(cls, mutation, value, story, operator):
```
judepereira: @stevepeak Running this file through pep8 produces the following warning:
```
asyncy/processing/Mutations.py:28:5: C901 'Mutations._mutate_list' is too complex (21)
```

What's the solution here? The current complexity is set to 15. Should we just increase it to 25?

asyncy/processing/Mutations.py
stevepeak: Same as map in python

```python
map(func, [1, 2, 3])
```

asyncy/processing/Mutations.py
stevepeak: Same as Python `reduce(func, list)` -- http://book.pythontips.com/en/latest/map_filter.html

asyncy/processing/Mutations.py
stevepeak: Return a list of `[b, d]`.
> Use keys for list of `[a, c]`

asyncy/processing/Mutations.py
stevepeak: Not sure TBO. @Vesuvium thoughts?

asyncy/processing/Mutations.py
judepereira: Should we leave this for later? SS doesn't support inline function or passing function references yet.

asyncy/processing/Mutations.py
judepereira: Should we leave this for later? Same argument as the one for `apply`.

asyncy/processing/Mutations.py
judepereira: Isn't it the same as `values` just above it? IMHO `keys` and `values` are self explanatory unlike items.

stevepeak: Oh, maybe items is `[[a, b], [c, d]]`?

asyncy/processing/Mutations.py
stevepeak: It behaves different though... it fold the results into the function instead of running through them one-by-one.

asyncy/processing/Mutations.py
stevepeak: Yea, that's ok

asyncy/processing/Mutations.py
judepereira: Yea, so items is this:
```
dict_items([('a', 1), ('b', 2)])
```
Given that we don't support this type (tuple) in SS yet, let's postpone this.

asyncy/processing/Mutations.py
Vesuvium: Metaprogramming:

```python
def _mutate_list(cls, mutation, value, story, operator):
    return getattr(ConvenientMutationsObject, operator)(*args)
```

In this way each mutation will correspond to a function with the same name of the mutation. In this way you can also reuse mutations that take different types but are the same code. You may find infact that it may be easier to pass mutations straight away to these functions and give them the ability to deal with different types.

asyncy/processing/Mutations.py
```diff
@@ -0,0 +1,137 @@
...
+        if operator == 'is_odd':
+            return value % 2 == 1
+        elif operator == 'is_even':
+            return value % 2 == 0
+        elif operator == 'absolute':
+            return abs(value)
+        elif operator == 'decrement':
+            return value - 1
+        elif operator == 'increment':
+            return value + 1
```
stevepeak: Decrement and increment are self mutating actions, including:
- `list sort`
- `map pop`
- `map apply`
- `list add`
- `list remove`
- `list reverse`
- `map update`
- `map shift`

tests/unit/processing/Mutations.py
```diff
@@ -0,0 +1,163 @@
...
+    # case[2] = operand (RHS operand. If not required, pass None)
+    # case[3] = expected return type
+    # case[4] = inplace (True if case[0] changes - inplace update)
+
+    # --- BEGIN - list ---
+    [['a', 's', 'y', 'n', 'c', 'y'], 'index', 's', 1, False],
+    [['a', 's', 'y', 'n', 'c', 'y'], 'index', 'z', -1, False],
+    [['a', 's', 'y', 'n', 'c', 'y'], 'length', None, 6, False],
+    [['a', 's', 'y', 'n', 'c', 'y'], 'join', '', 'asyncy', False],
+    [['a', 's', 'y', 'n', 'c', 'y'], 'join', ':', 'a:s:y:n:c:y', False],
```
stevepeak: 🎉

asyncy/processing/Mutations.py
judepereira: That did cross my mind, although I thought that a if/else would be faster.

Vesuvium: It is negligibly slower, and at this time performance is not the main issue

PR #120, opened 2018-08-02
Support for when statements

judepereira:


judepereira:
Covergage for Containers.py will be fixed with #126.

PR #138, opened 2018-09-05
Add cloud support

judepereira:


asyncy/Service.py
```diff
@@ -45,21 +45,10 @@ def main():
     @click.option('--debug',
                   help='Sets the engine into debug mode',
                   default=False)
-    @click.option('--sentry_dsn',
```
stevepeak: The SentryDNS and Release should still be used for starting the service. These are global to the Engine for tracking internal bugs (not user bugs).

judepereira: Got it.

PR #139, opened 2018-09-12
Attributes

judepereira:


judepereira:
Attributes don't really require any changes, as the response body syntax is a plain map.

PR #147, opened 2018-10-12
Support array access

judepereira:


judepereira:
Merging this despite of Codecov reports because:
1. Some statements do not compile in Storyscript, hence lowering the coverage
2. GraphQLAPI has a pending discussion regarding exposing app_hidden.repos.

PR #160, opened 2018-11-19
Create dev setup script

### LINKED ISSUE #153: Read local ~/.kube/config if the CLUSTER_* creds are missing
This will help with local development, and make it easier to get started.

First thought on approaching this - in Config.py, it reads the CLUSTER_* environment variables. After the apply() method has been called in __init__(), if the CLUSTER_* variables are null, attempt to read the local Kubernetes config file. Usage of pyyaml is expected (to parse and read).

Another thought would be to move over to the Kubernetes async Python client altogether, since this is pretty much duplicate code from that library.

rewop:
This PR creates a `asyncy-server dev_setup` script that creates needed resources in kubernetes, and creates a .env file with environment variables set.

What the script doesn't do yet:
- it doesn't initialize postgresql as for the database repo
- it doesn't accept arguments yet

resolve #153

### DIFF
--- a/asyncy/Service.py
+++ b/asyncy/Service.py
@@ -11,6 +11,7 @@ import prometheus_client
 import tornado
 from tornado import web
 from kubernetes import config as kube_config, client as kube_client
+import subprocess
 
 from . import Version
 from .Apps import Apps
@@ -91,23 +92,68 @@ class Service:
     @staticmethod
     @main.command()
     def dev_setup():
+        """Sets up development environment to run platform engine.
+        It creates a .env file with the value to be ued to run the engine.
+        This command assumes that kubectl is installed and connected to
+        a cluster and that postgres is installed and reachable through
+        localhost:5432
+        """
+
+        # load configuration from default
         kube_config.load_kube_config()
         kube_config_inst = kube_config.kube_config.Configuration()
 
+        # Create resources needed in the cluster
+        # @todo these resources will be created in the namespace default
+        # Should we also create a namespace for the engine?
+        kube_resources = [
+            'service_accounts/engine.yaml',
+            'secrets/engine.yaml',
+            'role_bindings/engine.yaml',
+        ]
+        command = ('curl -s '
+                   'https://raw.githubusercontent.com/asyncy/stack-kubernetes/'
+                   'master/kubernetes-pre-init/{resource}'
+                   ' | kubectl apply -f -')
+        for resource in kube_resources:
+            click.echo(f'creating or updating resource {resource}')
+            resp = subprocess.run(command.format(
+                resource=resource), shell=True)
+            resp.check_returncode()
+
         # CLUSTER_HOST
-        click.echo(f"CLUSTER_HOST={kube_config_inst.host}")
+        host = kube_config_inst.host
 
         # CLUSTER_CERT
         cert_file = open(kube_config_inst.cert_file, "r")
         cert = cert_file.read().replace("\n", "\\n")
-        click.echo(f"CLUSTER_CERT={cert}")
 
         # CLUSTER_AUTH_TOKEN
         v1Client = kube_client.CoreV1Api()
-        ret = v1Client.list_namespaced_secret(
+        res = v1Client.list_namespaced_secret(
             'default', pretty='true')
-        token = ret.items[0].data.get('token')
-        click.echo(f"CLUSTER_AUTH_TOKEN={token}")
+        secret = next(x for x in res.items
+                      if x.metadata.annotations
+                      .get('kubernetes.io/service-account.name') == 'engine')
+        token = secret.data.get('token')
+
+        # POSTGRES
+        # @todo we could parametrize these or adjust the
+        # defaults in config.py to use 'asyncy' db instead of
+        # 'postgres'
+        postgres_options = ('options=--search_path='
+                            'app_public,app_hidden,app_private,public '
+                            'dbname=asyncy '
+                            'user=postgres')
+
+        env_file = open('.env', 'w')
+        env_file.write(f'CLUSTER_HOST={host}\m')
+        env_file.write(f'CLUSTER_CERT={cert}\n')
+        env_file.write(f'CLUSTER_AUTH_TOKEN={token}\n')
+        env_file.write(f'POSTGRES={postgres_options}\n')
+        env_file.write(f'LOGGER_LEVEL=debug\n')
+
+        env_file.close()
 
     async def init_wrapper(sentry_dsn: str, release: str):
         try:
--- a/asyncy/Service.py
+++ b/asyncy/Service.py
@@ -91,12 +91,15 @@ class Service:
 
     @staticmethod
     @main.command()
-    def dev_setup():
+    @click.option('--psql-host',
+                  help='Hostname to connect to postgresql',
+                  default='localhost')
+    @click.option('--psql-port',
+                  help='Hostname to connect to postgresql',
+                  default='5432')
+    def dev_setup(psql_host, psql_port):
         """Sets up development environment to run platform engine.
         It creates a .env file with the value to be ued to run the engine.
-        This command assumes that kubectl is installed and connected to
-        a cluster and that postgres is installed and reachable through
-        localhost:5432
         """
 
         # load configuration from default
@@ -144,7 +147,9 @@ class Service:
         postgres_options = ('options=--search_path='
                             'app_public,app_hidden,app_private,public '
                             'dbname=asyncy '
-                            'user=postgres')
+                            'user=postgres '
+                            f'host={psql_host} '
+                            f'port={psql_port} ')
 
         env_file = open('.env', 'w')
         env_file.write(f'CLUSTER_HOST={host}\m')
--- a/asyncy/Service.py
+++ b/asyncy/Service.py
@@ -9,7 +9,6 @@ import click
 import prometheus_client
 
 import tornado
-from tornado import web
 from kubernetes import config as kube_config, client as kube_client
 import subprocess
 
@@ -172,7 +171,7 @@ class Service:
                             f'host={psql_host} '
                             f'port={psql_port} ')
 
-        env_file.write(f'CLUSTER_HOST={host}\m')
+        env_file.write(f'CLUSTER_HOST={host}\n')
         env_file.write(f'CLUSTER_CERT={cert}\n')
         env_file.write(f'CLUSTER_AUTH_TOKEN={token}\n')
         env_file.write(f'POSTGRES={postgres_options}\n')
--- a/asyncy/Service.py
+++ b/asyncy/Service.py
@@ -10,6 +10,7 @@ import prometheus_client
 
 import tornado
 from tornado import web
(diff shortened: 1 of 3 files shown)


asyncy/Service.py
```diff
@@ -88,6 +90,71 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
         logger.log_raw('info', 'Shutdown complete!')
 
     @staticmethod
+    @main.command()
+    def dev_setup():
+        """Sets up development environment to run platform engine.
+        It creates a .env file with the value to be ued to run the engine.
+        This command assumes that kubectl is installed and connected to
+        a cluster and that postgres is installed and reachable through
+        localhost:5432
```
judepereira: @rewop - Can you accept the host and port for this as a parameter?

asyncy/Service.py
```diff
@@ -88,6 +90,71 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
         logger.log_raw('info', 'Shutdown complete!')
 
     @staticmethod
+    @main.command()
+    def dev_setup():
+        """Sets up development environment to run platform engine.
+        It creates a .env file with the value to be ued to run the engine.
+        This command assumes that kubectl is installed and connected to
```
judepereira: Add a test for kubectl being installed and available.

asyncy/Service.py
```diff
@@ -88,6 +90,71 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+        a cluster and that postgres is installed and reachable through
+        localhost:5432
+        """
+
+        # load configuration from default
+        kube_config.load_kube_config()
+        kube_config_inst = kube_config.kube_config.Configuration()
+
+        # Create resources needed in the cluster
+        # @todo these resources will be created in the namespace default
```
judepereira: Make use of the namespace `asyncy-system`.

asyncy/Service.py
```diff
@@ -88,6 +90,71 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+        # Should we also create a namespace for the engine?
+        kube_resources = [
+            'service_accounts/engine.yaml',
+            'secrets/engine.yaml',
+            'role_bindings/engine.yaml',
+        ]
+        command = ('curl -s '
```
judepereira: Is it possible to use python `requests` to fetch the file, and then apply it with the `kubernetes` python library?

asyncy/Service.py
```diff
@@ -88,6 +90,71 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+                   'master/kubernetes-pre-init/{resource}'
+                   ' | kubectl apply -f -')
+        for resource in kube_resources:
+            click.echo(f'creating or updating resource {resource}')
+            resp = subprocess.run(command.format(
+                resource=resource), shell=True)
+            resp.check_returncode()
+
+        # CLUSTER_HOST
+        host = kube_config_inst.host
```
judepereira: Need to show a picker here before selecting the cluster to be used. Incase the cluster selected is some production cluster, we do not want to accidentally mess with it.

We need to make this CLI interactive, and ask for a Y/n explicitly.

asyncy/Service.py
```diff
@@ -88,6 +90,71 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+        # POSTGRES
+        # @todo we could parametrize these or adjust the
+        # defaults in config.py to use 'asyncy' db instead of
+        # 'postgres'
+        postgres_options = ('options=--search_path='
+                            'app_public,app_hidden,app_private,public '
+                            'dbname=asyncy '
+                            'user=postgres')
+
+        env_file = open('.env', 'w')
```
judepereira: Can you add the corresponding code to accept a path to an `env` file?

asyncy/Service.py
rewop: I tried, but I couldn't find any counterpart of `kubectl apply` command from the api. The `apply` command doesn't perform any action if the resources are not updated.

I could of course check if resources are already created, but I thought this was easier and cleaner. What do you think?

rewop:
@judepereira I addressed your comments. Could you check it again? Thanks

asyncy/Service.py
judepereira: Hmm, okay. In that case, we can just rely on `kubectl`.

asyncy/Service.py
```diff
@@ -88,6 +89,96 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+    @main.command()
+    @click.option('--psql-host',
+                  help='Hostname to connect to postgresql',
+                  default='localhost')
+    @click.option('--psql-port',
+                  help='Hostname to connect to postgresql',
+                  default='5432')
+    @click.argument('env_file', type=click.File('w'), default='.env')
+    def dev_setup(psql_host, psql_port, env_file):
+        """Sets up development environment to run platform engine.
```
judepereira:
```suggestion
        """
        Sets up development environment to run platform engine.
```

asyncy/Service.py
```diff
@@ -88,6 +89,96 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+        It creates a .env file with the value to be ued to run the engine.
```
judepereira:
```suggestion
        It creates a .env file with the value to be used to run the engine.
```

asyncy/Service.py
```diff
@@ -88,6 +89,96 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+        """
+
+        from shutil import which
+
+        # Make sure kubectl is installed
+        if which('kubectl') is None:
+            print(('Missing kubectl cli. Follow '
```
judepereira: Use click.echo().

asyncy/Service.py
```diff
@@ -88,6 +89,96 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+                   'to install it first.'), file=sys.stderr)
+            exit(1)
+
+        # Load configuration from default
+        kube_config.load_kube_config()
+        kube_config_inst = kube_config.kube_config.Configuration()
+
+        # CLUSTER_HOST
+        host = kube_config_inst.host
+        click.echo(f'ATTENTION: resources will be create in cluster at {host}')
```
judepereira:
```suggestion
        click.echo(f'ATTENTION: resources will be created in cluster at {host}')
```

asyncy/Service.py
```diff
@@ -88,6 +89,96 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+                   'master/kubernetes-pre-init/{resource}'
+                   ' | kubectl apply --namespace asyncy-system -f -')
+        for resource in kube_resources:
+            click.echo(f'creating or updating resource {resource}')
+            resp = subprocess.run(command.format(
+                resource=resource), shell=True)
+            resp.check_returncode()
+
+        # CLUSTER_CERT
+        cert_file = open(kube_config_inst.cert_file, "r")
```
judepereira:
```suggestion
        cert_file = open(kube_config_inst.cert_file, 'r')
```

asyncy/Service.py
```diff
@@ -88,6 +89,96 @@ def start(port, debug, sentry_dsn, release, prometheus_port):
...
+        cert = cert_file.read().replace("\n", "\\n")
```
judepereira:
```suggestion
        cert = cert_file.read().replace('\n', '\\n')
```

asyncy/Service.py
rewop: What I could do maybe is using the same approach I use for namespace. Create the resource and check for conflict error. Though i would have to request the yaml files and parse then, which would make the code a bit more verbose.

judepereira:
@rewop Nice work! 

What do you think about checking to see if the database is reachable and bootstrapped? And if not, print some message to clone asyncy/database, run the bootstrap. Thoughts on doing this?

This script looks almost complete.

**Some todos:**
- [ ] Move the dev command to another file alongside Service.py, say CliDevEnvHelper.py
- [ ] Write unit tests
- [ ] Ensure Circle CI tests succeed

asyncy/Service.py
judepereira: Naa, don't worry about it. It's much cleaner this way.

asyncy/Service.py
rewop: I use `print` to print on stderr. I haven't found a way to do so with click

rewop:
@judepereira thanks to check again.

> What do you think about checking to see if the database is reachable and bootstrapped? And if not, print some message to clone asyncy/database, run the bootstrap. Thoughts on doing this?

That could be done of course. I just have 2 thoughts:
1. what check would prove if the db is bootstrapped? 
2. I assume a developer when using this script is following a getting started before using this command, which should give the db bootstrapped. The goal of the script is to create the `.env` file in the end, and the db is not an hard requirement. Shouldn't a developer be skilled enough to have the db correctly set? 

About writing unit tests, I am not sure of the value of it. The command is just a sequence of I/O operations. I would be mainly mocking all those I/O and check they have been called, but this is done by just running the script. I am not sure the effort is worth the value. What do you think?

I will move the code in a different command.

rewop:
@judepereira can you give a look again? I fixed the errors from flake in CI though now there is a new error that looks like some network problem, and I cannot restart it.

I will wait your answer about my comment on test before proceeding with them.

asyncy/Service.py
judepereira: `click.echo('Hello World!', err=True)`

judepereira:
> what check would prove if the db is bootstrapped?

The table `releases` exists.

> I assume a developer when using this script is following a getting started before using this 
command, which should give the db bootstrapped. The goal of the script is to create the .env file in the end, and the db is not an hard requirement. Shouldn't a developer be skilled enough to have the db correctly set?

I agree, however, we must strive to make this bullet proof. Somebody who wants to just play around and somehow reaches here directly, shouldn't be frustrated (obviously, the error will make it obvious that the connection to the DB failed, but we should detect and prevent it if possible). Also note - the `.env` file contains PostgreSQL creds, hence we should validate that the engine can connect to it, and that it's bootstrapped.

> I fixed the errors from flake in CI though now there is a new error that looks like some network problem, and I cannot restart it

Yes, it did seem to be a one off issue. I just triggered it again, and it worked.

> About writing unit tests, I am not sure of the value of it. The command is just a sequence of I/O operations. I would be mainly mocking all those I/O and check they have been called, but this is done by just running the script. I am not sure the effort is worth the value. What do you think?

As of now, we cannot say with confidence that the script will remain the same throughout the lifetime of this project. Somebody is bound to contribute to it in the future, enhance it, or even change it significantly. Hence, to ensure that what was working continues to work when somebody else contributes, IMO asserting that the call sequence + what's being written to the `.env`
 is indeed the right output produced from the right command.

rewop:
@judepereira thanks to take time to respond. I will create tests. I will let you know when it is done.

judepereira:
@rewop Are tests the only thing pending here? Can I test drive this locally?

rewop:
Yes tests are the only part missing.

judepereira:
Closing as this PR is now stale.

PR #253, opened 2019-04-02
Add stack traces to Storyscript execution errors

### LINKED ISSUE #238: Produce stack traces for any errors during Storyscript execution
When an error occurs, there's no way to know in which story, or at which line it happened.

For example:
```coffee
http server as s
    when si listen path:'/fullcontact1' as req
        req write content:(fullcontact person email:req.query_params['email'])
```

The story about produces the following error:
```
Mar 04 17:04:16  ERROR The service "si:latest" was not found in the Asyncy Hub. Hint: 1. Check with the Asyncy team if this service has been made public; 2. Service names are case sensitive
```

A better message would be:
```
RuntimeException: The service "si:latest" was not found in the Asyncy Hub. Hint: 1. Check with the Asyncy team if this service has been made public; 2. Service names are case sensitive
        in hello.story:2
        in hello.story:1
```

The structure for this already exists - AsyncyError is thrown when there's a user related error. It also contains the `story` object, and the `line` object. Prettifying it can be done somewhere in here: 
https://github.com/asyncy/platform-engine/blob/master/asyncy/Logger.py#L24

bokan:
Closes #238


asyncy/Exceptions.py
```diff
@@ -6,7 +6,16 @@ class AsyncyError(Exception):
     def __init__(self, message=None, story=None, line=None):
         self.story = story
         self.line = line
-        super().__init__(message)
+        super().__init__(f'{self.__class__.__name__}: '
+                         f'{message}{self.trace(line)}')
```
judepereira: Calling trace directly won't work. This is because story and line are now set in a differed fashion. This is done because deep under the hood, `story` and `line` aren't always available. Hence, these two parameters are set by the original line executor (the only person who knows how to execute a line, therefore, a guarantee that `story` and `line` will be set eventually).

https://github.com/asyncy/platform-engine/blob/d0592b4fe2faef3bc1bb7b0502ec3c0a30b11afa/asyncy/processing/Story.py#L82-L85

asyncy/Exceptions.py
```diff
@@ -6,7 +6,16 @@ class AsyncyError(Exception):
...
+
+    def trace(self, line, stack=''):
+        if line and 'ln' in line:
+            stack = f'\n\tin {self.story.name}:{line["ln"]}'
```
wilzbach: A line number could be `1.4` (the . signifies compiler generated lines). This part should probably be hidden from the user and only the main line be displayed, e.g. `line.split('.')[0]`

judepereira: @wilzbach Can you start including the raw Storyscript lines so that this can be made better? Currently, it only prints the line number.

wilzbach: Uhm, sure, but for the compiler generated lines this will require an AST formatter (not hard to do, but required to go back to the actual expression). It might be a bit tricky with string interpolation and other shenanigans, but I guess we can just print the lines when known.

judepereira: > string interpolation and other shenanigans

LOL.

wilzbach: A first start: https://github.com/storyscript/storyscript/pull/774

wilzbach: This is now released as 0.16.0: https://github.com/storyscript/storyscript/releases/tag/0.16.0

judepereira:
Closing in favour of #344.

PR #257, opened 2019-04-09
Upgrade Storyscript to 0.14.3

### LINKED ISSUE #255: Upgrade Storyscript to 0.14.1
The unit tests in Resolver.py which compile Storyscript lines and calls the resolver to test them fail. This is because Storyscript now checks for variables which are used without any initialisation. Resolver.py used to set variables in the context, and then compile a Storyscript line directly and execute it in a haphazard manner to test the actual Resolver:
https://github.com/asyncy/platform-engine/blob/1c1ca6d7c24dfba3aeccf0ce71df0c81e4a9651b/tests/unit/utils/Resolver.py#L48-L77

However, since then, the engine has a new way of running actual Storyscript as integration tests, with various types of assertions:
https://github.com/asyncy/platform-engine/blob/491fcac1fcd3168e4025877745e111414c6100e5/tests/integration/processing/Lexicon.py#L29

The way these run are explained here:
https://github.com/asyncy/platform-engine/blob/491fcac1fcd3168e4025877745e111414c6100e5/tests/integration/processing/Lexicon.py#L436-L447

anukul:
Closes #255

### DIFF
--- a/tests/unit/utils/Resolver.py
+++ b/tests/unit/utils/Resolver.py
@@ -1,93 +1,10 @@
 # -*- coding: utf-8 -*-
-import re
-
 from asyncy.utils import Resolver
 
 import pytest
-from pytest import mark
-
-import storyscript
-
-
-@mark.parametrize('case', [
-    ['2 + 2', 4],
-    ['0 + 2', 2],
-    ['2 / 2', 1],
-    ['"a" + "b"', 'ab'],
-    ['"a" + "b" + ("c" + "d")', 'abcd'],
-    ['2 + 10 / 5', 4],
-    ['20 * 100', 2000],
-    ['10 % 2', 0],
-    ['11 % 2', 1],
-    ['20 / 1000', 0.02],
-    ['2 * 4 / 4', 2],
-    ['2 * 4 / (4 * 2)', 1],
-    ['2 * 4 / (4 * 2) + 1', 2],
-    ['2 * 4 / (4 * 2) + 1 * 20', 21],
-    ['2 * 4 / (4 * foo) + 1 * 20 + zero', 21, {'foo': 2, 'zero': 0}],
-])
-def test_resolve_expressions(case):
-    # case[0] == the expression under test
-    # case[1] == the expected result
-    # case[2] == context
-    t = storyscript.Api.loads(f'a = {case[0]}')
-    context = {}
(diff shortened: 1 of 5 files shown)


wilzbach:
0.14.2 is released by now.

anukul:
Ready for review!

tests/unit/utils/Resolver.py
```diff
@@ -44,50 +199,238 @@ def test_expression_invalid_type():
...
+                          'if !i\n'
+                          '    success = true',
+        cases=[
+            TestCase(assertion=ContextAssertion(key='success', expected=True))
+        ]
+    ),
+    TestSuite(
+        preparation_lines='i = null\n'
+                          'success = true\n'
+                          'if !i\n'
```
anukul: !null is **not** true?

tests/unit/utils/Resolver.py
```diff
@@ -44,50 +199,238 @@ def test_expression_invalid_type():
...
-    [{'i': 5}, 'if i < 5\n  echo foo', False, False],
-    [{'i': 5}, 'if i < 6\n  echo foo', True, False],
-    [{'i': 5}, 'if i <= 5\n  echo foo', True, False],
-    [{'i': 5}, 'if i <= 4\n  echo foo', False, False],
-    [{'i': 5}, 'if i <= 6\n  echo foo', True, False],
-    [{'i': False}, 'if i\n  echo foo', False, False],
-    [{'i': True}, 'if i\n  echo foo', True, False],
-    [{'i': True}, 'if !i\n  echo foo', False, False],
-    [{'i': False}, 'if !i\n  echo foo', True, False],
-    [{'i': None}, 'if !i\n  echo foo', False, False],
```
anukul: !null is **not** true?

judepereira: In this case, null is treated as false.

anukul: So if `null` is `false`, then `!null` should be `true`, right?

judepereira:
@anukul See the Codecov report - it's reporting lesser coverage, when there should've ideally been more.

tests/unit/utils/Resolver.py
judepereira: Yes.

PR #262, opened 2019-04-15
Update Storyscript to 0.16.1

judepereira:


asyncy/Stories.py
```diff
@@ -216,14 +216,15 @@ def function_line_by_name(self, function_name):
         return self.line(line_number)
 
     def argument_by_name(self, line, argument_name, encode=False):
-        args = line.get('args', line.get('arguments'))
+        args = line.get('args', line.get('arguments', line.get('arg')))
```
wilzbach: It's just `args` now.

asyncy/utils/Resolver.py
```diff
@@ -109,7 +111,7 @@ def expression(cls, item, data):
 
         left = cls.resolve(values[0], data)
 
-        if a == 'equals':
+        if a == 'equals' or a == 'equal':
```
wilzbach: Why keep the old name?

asyncy/utils/Dict.py
```diff
@@ -12,15 +12,26 @@ def set(_dict, keys, output):
             last = keys.pop()
             for key in keys:
                 if isinstance(_cur, list):
-                    _cur = _cur[int(key)]
+                    _cur = _cur[Dict.parse_int(key)]
```
wilzbach: It will always be an `$OBJECT` now, no?

PR #274, opened 2019-04-17
Format logs as JSON

anukul:


asyncy/Logger.py
```diff
@@ -94,6 +96,14 @@ def start(self):
         for event in self.events:
             self.frustum.register_event(event[0], event[1], event[2])
         self.frustum.start_logger()
+        self.set_formatter()
```
wilzbach: How about using plain logs when running the testsuite? I typically find them a lot easier to read for humans and most CI providers have no support for JSON at all.

judepereira:
Good idea. @anukul - can we have an environment variable that flips the format between the current and the new one? 


Cheers,
Jude Pereira

anukul:
Added.
Btw I haven't changed anything in the existing (text) format. Let me know if we want to get rid of the log level.
Coverage has reduced for test file, as expected.

asyncy/Logger.py
```diff
@@ -94,6 +98,16 @@ def start(self):
         for event in self.events:
             self.frustum.register_event(event[0], event[1], event[2])
         self.frustum.start_logger()
+        log_json = strtobool(os.getenv('LOG_JSON', 'False'))
```
judepereira:
```suggestion
        log_json = strtobool(os.getenv('LOG_FORMAT_JSON', False))
```

setup.py
```diff
@@ -40,7 +40,8 @@
         'certifi>=2018.8.24',
         'psycopg2==2.7.5',
         'requests==2.21.0',  # Used for structures like CaseInsensitiveDict.
-        'google-cloud-logging==1.9.0'
+        'google-cloud-logging==1.9.0',
+        'python-json-logger==0.1.11'
```
judepereira: Umm, can't we do this without this lib?

tests/unit/Logger.py
```diff
@@ -142,12 +143,26 @@ def test_logger_adapter(patch, magic, logger):
     Adapter.__init__.assert_called_with(logger.frustum.logger, extra)
 
 
-def test_logger_start(patch, logger):
+@mark.parametrize('log_json', [True, False])
+def test_logger_start(patch, logger, log_json):
     patch.many(Frustum, ['register_event', 'start_logger'])
+    patch.object(logger, 'set_json_formatter')
+    patch.dict(os.environ, {'LOG_JSON': str(log_json)})
```
judepereira:
```suggestion
    patch.dict(os.environ, {'LOG_FORMAT_JSON': str(log_json)})
```

asyncy/Logger.py
anukul: Changed the env var, left the `'False'` unchanged since `strtobool` needs a string

judepereira: Got it.

judepereira:
Closed, since your branch has been merged with platform-engine/anukul-logging

PR #279, opened 2019-04-24
Lexicon: Add support for subtraction

### LINKED ISSUE #273: Subtraction isn't implemented
```coffee
a = 1 - 2
```

yields:

<details>

```json
{
  "tree": {
    "1": {
      "method": "expression",
      "ln": "1",
      "name": [
        "a"
      ],
      "args": [
        {
          "$OBJECT": "expression",
          "expression": "subtraction",
          "values": [
            {
              "$OBJECT": "int",
              "int": 1
            },
            {
              "$OBJECT": "int",
              "int": 2
            }
          ]
        }
      ],
      "src": "a = 1 - 2"
    }
  },
  "entrypoint": "1",
  "version": "0.17.3-dirty"
}
```

</details>

assimovt:
Fixes #273

### DIFF
--- a/asyncy/utils/Resolver.py
+++ b/asyncy/utils/Resolver.py
@@ -102,6 +102,7 @@ class Resolver:
         - not
         - or
         - sum
+        - subtraction
         - division
         - multiplication
         """
@@ -168,6 +169,11 @@ class Resolver:
                 result += r
 
             return result
+        elif a == 'subtraction':
+            right = cls.resolve(values[1], data)
+            assert type(left) in (int, float)
+            assert type(right) in (int, float)
+            return left - right
         elif a == 'multiplication':
             right = cls.resolve(values[1], data)
             assert type(left) in (int, float, str)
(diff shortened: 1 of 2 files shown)


asyncy/utils/Resolver.py
```diff
@@ -168,6 +169,11 @@ def expression(cls, item, data):
                 result += r
 
             return result
+        elif a == 'subtraction':
+            right = cls.resolve(values[1], data)
+            assert type(left) in (int, float)
```
wilzbach: Storyscript supports implicit casts of boolean to int (like e.g. Python does too). We never decided whether we want the compiler to insert casts or the engine to support it, but in the latter case the at assert would need to be expanded.

assimovt: @wilzbach gotcha, thanks for clarifying. shall we open a separate issue to support that for other ops as well?

wilzbach: Well, we first need to decide whether the engine or the SS compiler is supposed to do the casting, but I guess as long as we have any, the following is possible:

```coffee
a = {}  # Map[any,any]
a[0] = true
b = a[0] # any
c = b + 1 # 2
```

judepereira:
@wilzbach what's the status of this PR? Can we merge it?

PR #291, opened 2019-05-03
Deploy apps in parallel

### LINKED ISSUE #290: Deploy all apps in parallel during a restart
Right now, a restart of the engine takes forever to boot up, since it deploys apps sequentially. However, we can parallelise this to deploy 100 apps concurrently.

anukul:
Closes #290 

- Apps are now deployed in parallel during a restart
- All app deployments time out after 5 minutes

### DIFF
--- a/asyncy/Apps.py
+++ b/asyncy/Apps.py
@@ -171,9 +171,10 @@ class Apps:
                              daemon=True)
         t.start()
 
-        for release in releases:
-            app_id = release[0]
-            await cls.reload_app(config, glogger, app_id)
+        await asyncio.gather(*[
+            cls.reload_app(config, glogger, release[0])
+            for release in releases
+        ])
 
     @classmethod
     def get(cls, app_id: str):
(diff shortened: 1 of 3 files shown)


judepereira:
Very cool @anukul! Just one thing - can we limit parallel deployments to 100? The reasoning behind this is that if we deploy 1000 apps at once, we have no idea how the k8s API will respond to this.

anukul:
@judepereira updated. can you take a look at why the tests are failing?
i see
```
asyncy.Exceptions.ServiceNotFound: The service "asyncy/http:latest" was not found in the Asyncy Hub.
Hint: 1. Check with the Asyncy team if this service has been made public;
2. Service names are case sensitive
```

anukul:
@judepereira added the 100 limit to parallel deployments.
also updated a failing test because apparently the slug for `asyncy/http` changed to `storyscript/http`? the pull url is still the same though

PR #293, opened 2019-05-07
Update map assignment

### LINKED ISSUE #292: The engine doesn't support non-string map keys
The following is supported by SS, but it currently fails in the engine:

```coffee
a = {} # Map[any,any]
a[0] = 1
```

```coffee
a = {} as Map[int,int]
a[0] = 1
```

anukul:
Fixes #292

### DIFF
--- a/asyncy/utils/Dict.py
+++ b/asyncy/utils/Dict.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from .Resolver import Resolver
 
 
 class Dict:
@@ -19,18 +20,41 @@ class Dict:
             if isinstance(_cur, list):
                 _cur[Dict.parse_int(last)] = output
             else:
-                _cur[last] = output
+                _cur[Dict.parse_map_key(last, _dict)] = output
 
     @staticmethod
     def parse_int(s):
         if isinstance(s, str):
+            # backwards-compatibility
             return int(s)
         elif isinstance(s, int):
+            # general purpose dict
+            # ss output will never hit this branch
             return s
         elif isinstance(s, dict) and s.get('$OBJECT') == 'int':
             return s['int']
         else:
             raise Exception(f'Unable to parse {type(s)} as int.')
+
+    @staticmethod
+    def parse_map_key(item, context):
+        if isinstance(item, dict):
+            object_type = item.get('$OBJECT')
+            if object_type == 'string':
+                return item['string']
+            elif object_type == 'int':
+                return item['int']
+            elif object_type == 'float':
+                return item['float']
+            elif object_type == 'path':
+                return Resolver.path(item['paths'], context)
+        elif isinstance(item, str) and item.lstrip('+-').isdigit():
+            # backwards-compatibility
+            return int(item)
+        else:
+            # general purpose dict
+            # ss output will never hit this branch
+            return item
 
     @staticmethod
     def find(root, path, default_value=None):
--- a/tests/integration/processing/Lexicon.py
+++ b/tests/integration/processing/Lexicon.py
@@ -761,6 +761,28 @@ async def test_arrays(suite, logger):
             TestCase(assertion=ContextAssertion(key='a', expected=21))
         ]
     ),
+    TestSuite(
+        preparation_lines='a = {}\n',
+        cases=[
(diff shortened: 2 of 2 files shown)


asyncy/utils/Dict.py
```diff
@@ -16,10 +16,10 @@ def set(_dict, keys, output):
                 else:
                     _cur = _cur.setdefault(key, {})
 
-            if isinstance(_cur, list):
+            try:
                 _cur[Dict.parse_int(last)] = output
```
wilzbach: It could be a float too.

anukul:
Explanation for reduced coverage:
- looks like ss doesn't yet support float map keys
- for int map keys, integration tests don't cover because ss now outputs type aware objects.
unit tests use Dict.find(map, path) where path is a dot delimited string of map keys.
no way to specify integer keys here, unless we modify Dict.find
suggestions?

@wilzbach @judepereira

wilzbach:
> looks like ss doesn't yet support float map keys

Nah works fine:

```coffee
a = {2.3: 4.5}
```

yields:

<details>

```json
Symbol('tmp', Map[float,float])
{
  "tree": {
    "1": {
      "method": "expression",
      "ln": "1",
      "name": [
        "a"
      ],
      "args": [
        {
          "$OBJECT": "dict",
          "items": [
            [
              {
                "$OBJECT": "float",
                "float": 2.3
              },
              {
                "$OBJECT": "float",
                "float": 4.5
              }
            ]
          ]
        }
      ],
      "src": "a = {2.3: 4.5}"
    }
  },
  "entrypoint": "1",
  "version": "0.18.1-1-g9d6afc87-dirty"
}
```

</details>

anukul:
@wilzbach oh.. is setting float keys not supported. then?
```
a = {2.3: 4.5}
a[2.3] = 4.5
```

<details>

```
Error: syntax error in story at line 2, column 4

2|    a[2.3] = 4.5
         ^

E0041: `.` is not allowed here
```
</details>

wilzbach:
Oh that's a bug -> https://github.com/storyscript/storyscript/issues/877

But you can get around it for now by declaring a temporary variable ;-)

anukul:
That brings us to another issue, paths in map keys :P

wilzbach:
Yeah SS is in a very messy state, but please just open up issues at its repo and I will fix them by the end of the week.

anukul:
ah SS is fine in that regard, i was referring to adding that in the engine
but i guess that also forms part of this PR, so i did add it now.

@wilzbach @judepereira looks ready for review
the failing test is unrelated to this, i pushed a commit to fix that in another existing PR - https://github.com/storyscript/platform-engine/pull/291/commits/31b5719a76e82c092a36bfdf0fbffab7cb1b0096

wilzbach:
> oh.. is setting float keys not supported. then?
> Oh that's a bug -> storyscript/storyscript#877

FYI: Fixed in https://github.com/storyscript/storyscript/releases/tag/0.18.2

Please feel free to upgrade directly here. There are only two bug fixes in 0.18.2

wilzbach:
@anukul make sure to rebase against upstream/master ;-)

judepereira:
Ready to merge?

anukul:
@judepereira I think so. Can you check the coverage report? I'm confused :s

anukul:
@judepereira Looks good to me now.
Codecov was showing stale reports for some files (logging), so I just force pushed.
Also left in the `Dict.find` unit tests, but nuked all the rest as discussed.

judepereira:
@anukul Sorry, I forgot the initial discussion of why we had to remove the unit tests for `Dict.set`. 

Note: Story runtime is not the only user of this, but `Dict.set` is used in various places in code as a simple helper for accessing dictionaries.

anukul:
@judepereira Those tests were accessing dictionaries directly by string keys, while the story runtime only ever uses objects (`{ '$OBJECT': 'string', 'string': 'foo' }`) for that purpose, even in older versions.

I see two instances of the use of `Dict.set`, both of which are invoked to process story compilation output, so I was confused why it was supporting direct string key access.

Do I add them back in?

judepereira:
Yes, add them back.


Cheers,
Jude

anukul:
Done

PR #313, opened 2019-05-28
Cast null to null, instead of "None".

judepereira:
If a value is None, and you cast it to a string, it returns `'None'` - the string value, and not the actual None type.


tests/integration/processing/Lexicon.py
```diff
@@ -612,6 +612,8 @@ def __init__(self, cases, preparation_lines=''):
     all_lines = suite.preparation_lines
 
     if case.append is not None:
+        if case.append == 'c = a[0]["id"] as string':
+            print('found')
```
wilzbach: Debugging leftover

judepereira: Removed, and merged.

PR #317, opened 2019-05-31
Check for image pull errors during deployment

### LINKED ISSUE #286: Gracefully handle "ImagePull" errors when starting a container
Right now, the engine just freezes indefinitely (as it waits forever for k8s to start the container). However, the k8s error api should also be consumed, to know if the container is unschedule-able.

anukul:
Closes #286

### DIFF
--- a/asyncy/Kubernetes.py
+++ b/asyncy/Kubernetes.py
@@ -5,6 +5,7 @@ import json
 import ssl
 import time
 import typing
+import urllib.parse
 from asyncio import TimeoutError
 
 from tornado.httpclient import AsyncHTTPClient, HTTPResponse
@@ -415,6 +416,24 @@ class Kubernetes:
         return False
 
     @classmethod
+    async def check_for_image_errors(cls, app, container_name):
+        prefix = cls._get_api_path_prefix('pods')
+        qs = urllib.parse.urlencode({
+            'labelSelector': f'app={container_name}'
+        })
+        res = await cls.make_k8s_call(app, f'{prefix}/{app.app_id}/pods?{qs}')
+        cls.raise_if_not_2xx(res)
+        body = json.loads(res.body, encoding='utf-8')
+        for pod in body['items']:
+            for container_status in pod['status'].get('containerStatuses', []):
+                is_waiting = container_status['state'].get('waiting', False)
+                if is_waiting and is_waiting['reason'] == 'ImagePullBackOff':
+                    raise K8sError(
+                        message=f'Failed to pull image '
+                        f"'{container_status['image']}'"
+                    )
+
+    @classmethod
     async def create_deployment(cls, app, service_name: str, image: str,
                                 container_name: str, start_command: [] or str,
                                 shutdown_command: [] or str, env: dict,
@@ -549,6 +568,7 @@ class Kubernetes:
             if body['status'].get('readyReplicas', 0) > 0:
                 break
 
+            await cls.check_for_image_errors(app, container_name)
             await asyncio.sleep(1)
 
         app.logger.debug('Deployment is ready')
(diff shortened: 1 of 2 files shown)


anukul:
@judepereira Added the suggested changes.

asyncy/Kubernetes.py
```diff
@@ -414,6 +414,23 @@ def format_ports(cls, ports: {int}):
 
         return False
 
+    @classmethod
+    async def check_imagepullbackoff(cls, app, container_name):
+        prefix = cls._get_api_path_prefix('pods')
+        res = await cls.make_k8s_call(app,
+                                      f'{prefix}/{app.app_id}/pods?'
+                                      f'labelSelector=app={container_name}')
```
judepereira: does this work? Shouldn't the value be URL encoded?

asyncy/Kubernetes.py
```diff
@@ -414,6 +414,23 @@ def format_ports(cls, ports: {int}):
...
+        cls.raise_if_not_2xx(res)
+        body = json.loads(res.body, encoding='utf-8')
+        for pod in body['items']:
```
judepereira: assert that only one item is returned, because if multiple are returned, there's a problem somewhere. raise a K8sException.

asyncy/Kubernetes.py
```diff
@@ -414,6 +414,23 @@ def format_ports(cls, ports: {int}):
...
+            for container_status in pod['status'].get('containerStatuses', []):
+                is_waiting = container_status['state'].get('waiting', False)
+                if is_waiting and is_waiting['reason'] == 'ImagePullBackOff':
```
judepereira: what are the other types of image pull errors?

asyncy/Kubernetes.py
```diff
@@ -549,6 +566,7 @@ def format_ports(cls, ports: {int}):
             if body['status'].get('readyReplicas', 0) > 0:
                 break
 
+            await cls.check_imagepullbackoff(app, container_name)
```
judepereira: Please rename to `check_for_image_errors`.

asyncy/Kubernetes.py
anukul: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/images/types.go#L26-L44

I tried using an invalid host, invalid repository, invalid image, invalid credentials - all of which resulted in `ErrImagePull` and then `ImagePullBackOff`

asyncy/Kubernetes.py
anukul: Done

asyncy/Kubernetes.py
anukul: Upon a restart, I see two pods - the old one terminating and the new one starting up.

asyncy/Kubernetes.py
anukul: It works, but yeah I urlencoded the value now.

asyncy/Kubernetes.py
anukul: Failing on all image pull errors now, as discussed.

asyncy/Kubernetes.py
anukul: @judepereira It seems like there's no way to tell from this API call that a container is terminating.
See response [here](https://jsonblob.com/9ea8736e-8d5f-11e9-8bcb-8d3f0ea1e982). The first item in the array is the terminating pod, second is the new one - which as you can see has the `ContainerCreating` status.

It seems kind of out of flow to separately check (maybe by another API call) and confirm by pod name that the first one is terminating.

anukul:
@judepereira Looks good to me now.

PR #318, opened 2019-06-02
Use ImagePullSecrets

anukul:
Related:

- https://github.com/storyscript/platform-engine/issues/299
- https://github.com/storyscript/database/pull/72
- https://github.com/storyscript/cli/pull/177


asyncy/Containers.py
```diff
@@ -254,6 +291,10 @@ def hash_volume_name(cls, app, line, service, volume_name):
         h = hashlib.sha1(key.encode('utf-8')).hexdigest()
         return f'{simple_name}-{h}'
 
+    @classmethod
+    def new_pg_conn(cls, config: Config):
```
anukul: This function is a duplicate of `Apps.new_pg_conn` (didn't import due to a cyclic dependency between Apps and Containers).
Before this, all database operations were done from Apps.
Should it be moved somewhere different to keep one copy and import in both places, or is this cool?

asyncy/Containers.py
```diff
@@ -119,6 +125,37 @@ def is_service_reusable(cls, app, line):
                                            service_alias)
         return Kubernetes.get_hostname(story.app, container)
 
+    @classmethod
+    def get_registry_url(cls, image):
```
anukul: Taken from https://github.com/docker/cli/blob/c02f389c787f9c74a517c0f718ca05ea4ba886cd/vendor/github.com/docker/distribution/reference/normalize.go#L91
and
https://github.com/docker/cli/blob/c9d0e4741481c98c1aba337ce377d205b8905c4c/cli/command/registry.go#L80

asyncy/Containers.py
anukul: @judepereira Moved this function to `asyncy.Database`, didn't refactor anything else since this solved the cyclic dependency problem.

Let me know if anything else needs to be changed along with this.

asyncy/Containers.py
```diff
@@ -119,6 +123,35 @@ def is_service_reusable(cls, app, line):
...
+        else:
+            return image[:i]
+
+    @classmethod
+    async def get_docker_configs(cls, app, image):
+        registry_url = cls.get_registry_url(image)
+        conn = Database.new_pg_conn(app.config)
+        cur = conn.cursor()
+        query = f"""
+        with containerconfigs as (select name, owner_uuid, containerconfig,
```
anukul: Just noticed, we can make this into a GraphQL query.
Postgraphile [supports](https://github.com/graphile-contrib/postgraphile-plugin-connection-filter#json-jsonb) querying jsonb in the form we're doing here (containsKey)

asyncy/Apps.py
```diff
@@ -266,56 +242,36 @@ def get(cls, app_id: str):
...
-            version = release[1]
-            environment = release[2]
-            stories = release[3]
-            maintenance = release[4]
-            app_dns = release[5]
-            state = release[6]
-            deleted = release[7]
-            owner_uuid = release[8]
-            if state == ReleaseState.FAILED.value:
+            release = Database.get_release_for_deployment(config, app_id)
```
judepereira: Can we use an object here instead of a map? That way, it'll be far easier to update and use (since there is no guess work involved in what keys are there and what aren't).

asyncy/Apps.py
```diff
@@ -333,12 +289,11 @@ def get(cls, app_id: str):
     @classmethod
     def listen_to_releases(cls, config: Config, glogger: Logger, loop):
         glogger.info('Listening for new releases...')
-        conn = cls.new_pg_conn(config)
+        conn, cur = Database.new_pg_conn(config)
```
judepereira: Please don't use multi return.

asyncy/Kubernetes.py
```diff
@@ -400,6 +405,33 @@ def format_ports(cls, ports: {int}):
...
+        b64_docker_config = base64.b64encode(
+            json.dumps(config['dockerconfig']).encode()
+        ).decode()
+
+        payload = {
+            'apiVersion': 'v1',
+            'kind': 'Secret',
+            'type': 'kubernetes.io/dockerconfigjson',
+            'metadata': {
+                'name': config['name'],
```
judepereira: We need to normalise this name. The user (via the CLI) might include a special character. So, let's either hash it or normalise it. I recommend normalise + hash - just like the pod names.

asyncy/Containers.py
```diff
@@ -207,6 +212,12 @@ def format_command(cls, story, line, container_name, command):
 
         return command_parts
 
+    @classmethod
+    def get_dockerconfig_name(cls, name, config):
```
judepereira: Use `get_container_config_name` please.

asyncy/Containers.py
```diff
@@ -269,6 +280,11 @@ def hash_volume_name(cls, app, line, service, volume_name):
         h = hashlib.sha1(key.encode('utf-8')).hexdigest()
         return f'{simple_name}-{h}'
 
+    @classmethod
+    def hash_dockerconfig_name(cls, name, config):
```
judepereira: `hash_container_config_name`.

asyncy/Containers.py
```diff
@@ -269,6 +280,11 @@ def hash_volume_name(cls, app, line, service, volume_name):
...
+        return hashlib.sha1(f'{name}-{ujson.dumps(config)}'
```
judepereira: LOL - you're hashing the entire config as well with the name. Maybe hash the name and app version?

asyncy/Apps.py
```diff
@@ -242,35 +243,40 @@ def get(cls, app_id: str):
                 glogger.warn(f'Another deployment for app {app_id} is in '
                              f'progress. Will not reload.')
                 return
-            release = Database.get_release_for_deployment(config, app_id)
-            if release['state'] == ReleaseState.FAILED.value:
+            data = Database.get_release_for_deployment(config, app_id)
```
judepereira: Can you make the function `Database.get_release_for_deployment` return an object of `Release` itself? This way, somebody else using the same won't need to create the plain old Python object (POPO) outside.

asyncy/Kubernetes.py
```diff
@@ -408,8 +408,8 @@ def format_ports(cls, ports: {int}):
     @classmethod
     async def create_imagepullsecret(cls, app, config: dict):
```
judepereira: Is the `config` parameter something directly from the database? From the looks of it, I think it is. If this is the case, let's use a POPO (plain old Python object)?

asyncy/Database.py
```diff
@@ -39,7 +40,7 @@ def update_release_state(cls, glogger, config, app_id, version,
         glogger.info(f'Updated state for {app_id}@{version} to {state.name}')
 
     @classmethod
-    def get_docker_configs(cls, app, registry_url):
+    def get_container_configs(cls, app, registry_url):
         conn = cls.new_pg_conn(app.config)
         cur = conn.cursor(cursor_factory=RealDictCursor)
         query = f"""
```
judepereira: Don't use f-strings for database queries, as it opens us up to SQL injection. Use prepared statements. Look at the way the other queries are written (for releases, apps, etc). Hint: Use `%s`.

PR #344, opened 2019-06-14
Add support for stack traces + call stack

judepereira:


judepereira:
@wilzbach If you can run a first pass review, it'd be great. I've yet to add tests and fix the listing etc, but the code (apart from tests) is tested locally and runs fine.

Things to focus on:
1. Introduced a call stack (right now it's just line numbers in the stack, but later we can add context etc)
2. Simplified the stack trace approach (the approach earlier was a split approach between the engine and the CLI, making it difficult to maintain and push updates in the future)

judepereira:
This is what the current output looks like:
```
ERROR:storyscript:app_id::43 => An exception has occurred:
List index out of bounds: 1000
    at line 9: when server listen path: "/" as req (in /Users/jude/Library/Preferences/IntelliJIdea2019.1/scratches/scratch_33.txt)
    at line 11.1: method=call (auto generated frame) (in /Users/jude/Library/Preferences/IntelliJIdea2019.1/scratches/scratch_33.txt)
    at line 6: return a[1000] as int (in /Users/jude/Library/Preferences/IntelliJIdea2019.1/scratches/scratch_33.txt)
```

asyncy/Exceptions.py
```diff
@@ -3,11 +3,35 @@
...
+        for item in stack:
+            line = self.story.line(item)
+            src = line["src"]
+
+            if src is None:
+                src = f'method={line["method"]} (auto generated frame)'
+            
+            src = src.strip()
+
+            trace += f'\n    at line {item}: {src} (in {self.story.name})'
```
wilzbach: Should we really print line 11.10?

asyncy/processing/Lexicon.py
```diff
@@ -184,6 +185,7 @@ def _is_if_condition_true(story, line):
 
         # while true here because all if/elif/elif/else is executed here.
         while True:
+
```
wilzbach: unrelated

asyncy/processing/Story.py
```diff
@@ -51,8 +51,12 @@ def save(logger, story, start):
         :return: Returns the next line number to be executed
         (return value from Lexicon), or None if there is none.
         """
-        line = story.line(line_number)
+        line: dict = story.line(line_number)
         story.start_line(line_number)
+
+        ex_occurred = False
+        story.push_line_number_on_stack(line_number)
```
wilzbach: I recommend doing sth. like:

```
with story.add_stackframe(linenumber) as current_frame:
```

tests/integration/conftest.py
```diff
@@ -26,18 +26,56 @@ def app(magic):
...
+    # assert 0
     return Stories(app, 'hello.story', logger)
 
 
+@fixture
+def raw_stories(app, logger):
+    asset_dir = examples.__path__[0]
+
+    with open(asset_dir + '/stories.json', 'r') as file:
+        return ujson.loads(file.read())
```
wilzbach: This is wasteful. `services` should be used.

tests/integration/processing/Lexicon.py
```diff
@@ -100,12 +100,14 @@ def __init__(self, cases, preparation_lines=''):
         preparation_lines='function echo i:int returns int\n'
                           '    return i\n'
                           '\n'
+                          'x = 500\n'
```
wilzbach: unrelated change.

asyncy/Stories.py
```diff
@@ -25,9 +25,19 @@ def __init__(self, app, story_name, logger):
...
         self.execution_id = str(uuid.uuid4())
         self._tmp_dir_created = False
 
+    def push_line_number_on_stack(self, line_number: str):
+        self._stack.append(line_number)
+
+    def get_stack(self) -> []:
+        return self._stack
+
+    def pop_line_number_from_stack(self):
```
wilzbach: I highly recommend renaming this method as the stack is not only about line numbers.

asyncy/Stories.py
```diff
@@ -25,9 +25,19 @@ def __init__(self, app, story_name, logger):
         self.containers = None
         self.repository = None
         self.version = None
+        self._stack = []
         self.execution_id = str(uuid.uuid4())
         self._tmp_dir_created = False
 
+    def push_line_number_on_stack(self, line_number: str):
```
wilzbach: Don't you want to store e.g. the context variables too?

asyncy/Exceptions.py
judepereira: Yea, it kinda helps  for now. However, fake lines cause a small issue. We can deal with this later. 

An ideal solution would be to introduce a new files in the JSON tree which indicates the real line number in a Storyscript file (which is used for no other reason than for debugging purposes).

asyncy/processing/Lexicon.py
judepereira: Yes, I agree. Please ignore, since there was some intermediate change which I reverted here.

tests/integration/conftest.py
judepereira: Please ignore tests. I will refactor it all.

asyncy/processing/Story.py
judepereira: Excellent idea, I’ll do this if it’s quick and easy.

tests/integration/processing/Lexicon.py
judepereira: Agreed, ignore this (doesn’t do anything functional).

asyncy/Stories.py
judepereira: Will do.

asyncy/Stories.py
judepereira: Yes, I want to. However, I’ll do that in the next PR. I don't want to increase the scope of this PR as it has accidentally turned out to be large and contains some unrelated changes as well.

asyncy/processing/Story.py
judepereira: Done, will push in a bit. Adding some tests.

PR #345, opened 2019-06-15
Fix k8s_call in check_for_image_errors

anukul:


anukul:
First, the `check_for_image_errors` PR was merged, which was using the old `make_k8s_call` signature.
When the signature changed, this file was left unchanged because it was in a separate branch.

anukul:
Also pushed test to check `make_k8s_call` args from `check_for_image_errors` now

PR #346, opened 2019-06-15
Record resource usage metrics for services

anukul:


anukul:
Will change base to `master` after #318 is merged.

asyncy/ServiceUsage.py
```diff
@@ -0,0 +1,142 @@
...
+                pod['containers'][0]['usage']['memory'])
+
+        if num_pods != 0:
+            average_cpu = total_cpu / num_pods
+            average_memory = total_memory / num_pods
+
+        return average_cpu, average_memory, num_pods
+
+    @classmethod
+    async def record_service_usage(cls, config: Config, logger: Logger):
```
anukul: This has an infinite loop, how do I test it?
In the other infinite loop in the code (`listen_to_releases`) I noticed that it is made to break when unable to subscribe to the postgres notification, and that behavior is enforced in the test.

Any suggestions for this?

asyncy/ServiceUsage.py
```diff
@@ -0,0 +1,136 @@
...
+            'n': 1e-9,
+            'u': 1e-6,
+            # FYI: "precision finer than 1m is not allowed" in cpu limit input
+            'm': 1e-3,
+            'k': 1e3,
+            'M': 1e6,
+            'G': 1e9,
+            'T': 1e12,
+            'P': 1e15,
+            'E': 1e18
```
judepereira: I like the optimism here!

asyncy/ServiceUsage.py
judepereira: Add a flag `shutting_down` in `Service.py`. Then here, use `while not Service.shutting_down`.

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
...
+        # Store only the last ${METRICS_SIZE} metrics
+        data.update((k, v[-cls.METRICS_SIZE:]) for k, v in data.items())
+
+        conn, cur = cls.new_pg_conn(config)
+        query = f"""
+        update service_usage
+        set cpu_units = %s, memory_bytes = %s
+        where service_uuid = %s;
+        """
+        cur.execute(query, (*tuple(data.values()), service['uuid']))
```
judepereira: Simplify this line please. Use `data['cpu_units']`, `data['memory_bytes']`.

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
...
+        cur.execute(query, (service_alias,))
+        return cur.fetchone()
+
+    @classmethod
+    def get_service_by_slug(cls, config: Config,
+                            owner_username: str, service_name: str) -> Dict:
+        conn, cur = cls.new_pg_conn(config)
+        query = f"""
+        select services.uuid from services
+        join owners on owner_uuid = owners.uuid
```
judepereira: Maybe join on `app_hidden.repos` instead of `owners`?

asyncy/Kubernetes.py
```diff
@@ -530,6 +534,9 @@ def format_ports(cls, ports: {int}):
 
         b16_service_name = base64.b16encode(service_name.encode()).decode()
 
+        cpu_limit, memory_limit = Database.get_service_limits(app.config,
```
judepereira: Can you use an object instead of multi-return? We like readability :)

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
         """
         cur.execute(query, (app_id,))
         return cur.fetchone()
+
+    @classmethod
+    def get_all_services(cls, config: Config) -> List[Dict]:
+        conn, cur = cls.new_pg_conn(config)
```
wilzbach: Shouldn't we cache the database connections?

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
...
+        set cpu_units = %s, memory_bytes = %s
+        where service_uuid = %s;
+        """
+        cur.execute(query, (*tuple(data.values()), service['uuid']))
+        conn.commit()
+
+    @classmethod
+    def get_service_by_alias(cls, config: Config, service_alias) -> Dict:
+        conn, cur = cls.new_pg_conn(config)
+        query = f"""
```
wilzbach: No f-string.

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
...
+        select uuid from services where alias = %s
+        """
+        cur.execute(query, (service_alias,))
+        return cur.fetchone()
+
+    @classmethod
+    def get_service_by_slug(cls, config: Config,
+                            owner_username: str, service_name: str) -> Dict:
+        conn, cur = cls.new_pg_conn(config)
+        query = f"""
```
wilzbach:
```suggestion
        query = """
```

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
...
+        return cur.fetchone()
+
+    @classmethod
+    def get_service_limits(cls, config: Config, service: str):
+        if '/' in service:
+            service = cls.get_service_by_slug(config, *service.split('/'))
+        else:
+            service = cls.get_service_by_alias(config, service)
+        conn, cur = cls.new_pg_conn(config)
+        query = f"""
```
wilzbach:
```suggestion
        query = """
```

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
...
+        join owners on owner_uuid = owners.uuid
+        where owners.username = %s and services.name = %s
+        """
+        cur.execute(query, (owner_username, service_name))
+        return cur.fetchone()
+
+    @classmethod
+    def get_service_limits(cls, config: Config, service: str):
+        if '/' in service:
+            service = cls.get_service_by_slug(config, *service.split('/'))
```
wilzbach: Let's assert that we only have two pieces here before we do a spread.

asyncy/Database.py
```diff
@@ -71,3 +76,96 @@ def get_release_for_deployment(cls, config, app_id):
...
+        conn, cur = cls.new_pg_conn(config)
+        query = f"""
+        select cpu_units, memory_bytes
+        from service_usage
+        where service_uuid = %s
+        """
+        cur.execute(query, (service['uuid'],))
+        res = cur.fetchone()
+        if res is None or len(res['cpu_units']) < cls.METRICS_SIZE:
+            return '0', '200Mi'
```
wilzbach: Should we really use strings for the cpu_unit limit?

asyncy/Kubernetes.py
wilzbach: It also scales better for the future.

asyncy/ServiceUsage.py
```diff
@@ -0,0 +1,136 @@
...
+            if len(pod['containers']) > 1:
+                raise K8sError(
+                    message=f'Found {len(pod["containers"])} containers '
+                    f'in pod {pod["metadata"]["name"]}, expected 1')
+            # Convert metrics to standard units (no suffix)
+            total_cpu += cls.cpu_units(pod['containers'][0]['usage']['cpu'])
+            total_memory += cls.memory_bytes(
+                pod['containers'][0]['usage']['memory'])
+
+        if num_pods != 0:
```
wilzbach:
```suggestion
        if num_pods > 0:
```

asyncy/ServiceUsage.py
```diff
@@ -0,0 +1,136 @@
...
+    async def record_service_usage(cls, config: Config, logger: Logger):
+        while True:
+            all_services = Database.get_all_services(config)
+            for service in all_services:
+                # Get cpu, memory average for all running pods of the service
+                cpu_avg, memory_avg, num_pods = \
+                    await cls.get_pod_metrics(service, config, logger)
+                if num_pods == 0:
+                    # No running pods found, nothing to do here
+                    continue
```
wilzbach: You should sleep here too

asyncy/Database.py
anukul: I think so.
`0` just denotes an unbounded limit, but otherwise you'd generally use something like `500m` etc

asyncy/ServiceUsage.py
anukul: I guess not.. this skips to the next iteration of the `for` loop, updating metrics for the next service.
We're sleeping between the record jobs for all services.

asyncy/ServiceUsage.py
anukul: Weird import due to cyclic dependency (Service -> Apps -> ServiceUsage -> Service), please check if it looks okay.

asyncy/Database.py
anukul: Umm I'm not sure how the repo table is used in this context (or otherwise).
I wrote this query based on how the GraphQL query `get_by_slug` is set up to query by `owners.username` and `service.name`

anukul:
In the case that the first 25 measurements are too low, I guess we should leave some headroom for the limit - say 1.25x?

judepereira:
> In the case that the first 25 measurements are too low, I guess we should leave some headroom for the limit - say 1.25x?

@anukul yes, I like this idea.

asyncy/Database.py
anukul: Maybe in another PR for the new issue.

wilzbach: Yes I just point things out as I see them. For this I have already opened https://github.com/storyscript/platform-engine/issues/353, but in the future please feel free to open new issues for other things that you notice ;-)

asyncy/db/Database.py
```diff
@@ -86,3 +90,105 @@ def get_release_for_deployment(cls, config, app_id):
...
+    @classmethod
+    def update_service_usage(cls, config: Config, service, data):
+
+        # Store only the last ${METRICS_SIZE} metrics
+        data.update((k, v[-cls.METRICS_SIZE:]) for k, v in data.items())
+
+        with cls.new_pg_cur(config) as db:
+            query = """
+            update service_usage
+            set cpu_units = %s, memory_bytes = %s
```
wilzbach: Ouch so with every update we overwrite _all_ values (and have to send them to the DB)?
Two ideas:

1) Why not store the values in a separate table?

2) If it must be in one table. Postgres performs a lot better with fixed-size arrays. So why not use a fixed-size array on just update the current value of interest? This could be done by keeping the current index in the table and bumping it with every update.

asyncy/db/Database.py
```diff
@@ -86,3 +90,105 @@ def get_release_for_deployment(cls, config, app_id):
...
+
+    @classmethod
+    def update_service_usage(cls, config: Config, service, data):
+
+        # Store only the last ${METRICS_SIZE} metrics
+        data.update((k, v[-cls.METRICS_SIZE:]) for k, v in data.items())
+
+        with cls.new_pg_cur(config) as db:
+            query = """
+            update service_usage
```
wilzbach: Shouldn't we store the timestamp of the values (and their app version) as a new release of a service might perform entirely different?
Alternatively, of course, a new release would always need to purge all previous recorded usages...

asyncy/db/Database.py
```diff
@@ -86,3 +90,105 @@ def get_release_for_deployment(cls, config, app_id):
...
+            set cpu_units = %s, memory_bytes = %s
+            where service_uuid = %s;
```
wilzbach: Additionally, I should point out that storing log data in postgres is absolutely not what it's made for and will become a problem very soon. An explanation:

https://www.revsys.com/tidbits/three-things-you-should-never-put-your-database/

anukul:
What's the upper limit to this 1.25x scaling?

asyncy/ServiceUsage.py
```diff
@@ -0,0 +1,140 @@
...
+                if metrics['num_pods'] == 0:
+                    # No running pods found, nothing to do here
+                    continue
+                # Get stored metrics (past cpu, mem averages) of the service
+                usage = Database.get_service_usage(config, service)
+                # Add new set of entries to the usage arrays
+                usage['cpu_units'].append(metrics['average_cpu'])
+                usage['memory_bytes'].append(metrics['average_memory'])
+                Database.update_service_usage(config, service, usage)
+            await asyncio.sleep(cls.WAIT_PERIOD)
```
wilzbach: One thing that bothers me about the current implementation is that it scales poorly. We execute 2N database queries for n services. I think we can do a lot better.
e.g. for every 100 services do:

1) Run `get_pod_metrics` in parallel
2) Do a _single_ update

asyncy/db/Database.py
```diff
@@ -86,3 +90,105 @@ def get_release_for_deployment(cls, config, app_id):
...
+            res = db.cur.fetchone()
+            if res is None or len(res['cpu_units']) < cls.METRICS_SIZE:
+                limits = {
+                    'cpu': '0',
+                    'memory': '200Mi'
+                }
+            else:
+                limits = {
+                    'cpu': 1.25 * mean(res['cpu_units']),
+                    'memory': 1.25 * mean(res['memory_bytes'])
```
wilzbach: This is not very sophisticated. Let's at least do percentiles like e.g. the 95% percentile (~ two standard deviations from the mean)

https://en.wikipedia.org/wiki/Percentile

asyncy/db/Database.py
```diff
@@ -86,3 +90,105 @@ def get_release_for_deployment(cls, config, app_id):
...
+
+        with cls.new_pg_cur(config) as db:
+            query = """
+            select cpu_units, memory_bytes
+            from service_usage
+            where service_uuid = %s
+            """
+            db.cur.execute(query, (service['uuid'],))
+            res = db.cur.fetchone()
+            if res is None or len(res['cpu_units']) < cls.METRICS_SIZE:
```
wilzbach: what's the motivation for the second part (`len(res['cpu_units']) < cls.METRICS_SIZE`)? Why not use this info - even if it's just partially complete.

anukul:
@wilzbach This is decidedly a short term solution, the right way to do this would be using Prometheus (storage, advanced querying), although some of your comments still apply (parallel scraping, bulk update).

@judepereira Can you take a look and summarise what changes we need to make for now?

asyncy/db/Database.py
anukul: Because we can't decide the load to a decent extent in the very first measurement?

asyncy/db/Database.py
anukul: Because it's not a ton of values, just 25.. and it always stays that size.

asyncy/ServiceUsage.py
```diff
@@ -0,0 +1,140 @@
...
+                # Get cpu, memory average for all running pods of the service
+                metrics = await cls.get_pod_metrics(service, config, logger)
+                if metrics['num_pods'] == 0:
+                    # No running pods found, nothing to do here
+                    continue
+                # Get stored metrics (past cpu, mem averages) of the service
+                usage = Database.get_service_usage(config, service)
+                # Add new set of entries to the usage arrays
+                usage['cpu_units'].append(metrics['average_cpu'])
+                usage['memory_bytes'].append(metrics['average_memory'])
```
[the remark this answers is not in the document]
anukul: Yeah, it is fixed size. We can definitely discuss the limit calculation procedure in detail and come up with something better than mean.

asyncy/db/Database.py
wilzbach: Hmm, I'm not sure whether I follow, but isn't it better to have _some_ information and make a guess based on that than using the default?

anukul: Making a bad guess could cost us more than just allowing 25 data points to accumulate, which shouldn't take very long since they're app agnostic.

anukul: We can change the numbers (25, mean, 1.25) to something more appropriate.

asyncy/db/Database.py
wilzbach: > Because it's not a ton of values, just 25.. and it always stays that size.

Then why not use a fixed size array directly in the DB or just store the mean/deviation?

asyncy/Apps.py
```diff
@@ -156,6 +157,11 @@ def make_logger_for_app(cls, config, app_id, version):
                                             daemon=True)
         release_listener.start()
 
+        usage_recorder = threading.Thread(target=ServiceUsage.start_recording,
+                                          args=[config, glogger, loop],
+                                          daemon=True)
+        usage_recorder.start()
```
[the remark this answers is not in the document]
anukul: Anything in between this and Prometheus?

asyncy/ServiceUsage.py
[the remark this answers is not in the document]
wilzbach: The point here was that (1) if you're just interested in mean/stddev, you don't need to store any values.
And (2) as long as the recorder is running it never needs to get the old service usage data (it still has the fixed-size array in memory and no one else is updating it). In other words, you only need to do the `get_service_usage` on startup and then can just a ring buffer, no need to query the database constantly nor to always claim new memory from the Python GC.

asyncy/db/Database.py
wilzbach: > Making a bad guess could cost us more than

Hmm, so you're saying that an app might have a memory spike in the beginning and we record that and then it crashes. Thus, on the next start we would allocate more memory for it? In my experience, apps only continue to grow in memory-consumption and unfortunately they almost never get lighter. That being said, I agree that it shouldn't matter as in practice this shouldn't occur too often.

asyncy/db/Database.py
anukul: > Then why not use a fixed size array directly in the DB

Not sure how this works (truncation etc), any doc links?

> just store the mean/deviation

We should calculate the mean from the most recent K values.

wilzbach:
BTW @anukul do we also plan to dynamically down/up-size pods? If we do (which I hope is the case), we should talk a bit more about the recording of values as it matters then to quickly recognize trends.

asyncy/db/Database.py
wilzbach: > Not sure how this works (truncation etc), any doc links?

https://www.postgresql.org/docs/9.5/arrays.html

```sql
cpu_units float[25]
```

Currently, you update the entire array, the code wouldn't change.
If you only want to update one element, as explained above you could keep an index counter, though to be fair with 25 elements, that's probably not worth it.
While fixed-size arrays will make the DB a lot faster, they come with the downside that resizing them is a lot harder. So if this is really just experimental for now and we will be tweaking the value 25 over the next weeks, then better not use fixed for now.

> We should calculate the mean from the most recent K values.

Currently, every 5 minutes a timestamp is recorded. This means a full recording of 25 values takes 2 hours and 5 minutes. As explained somewhere else, as this is run by the same machine there's no need to continuously fetch the old values (they're still in memory after all and are guaranteed to be unchanged), so it wouldn't be too unreasonable to let the service recorder just write the `mean` into the database (or better the predicted pod size) as that's the only value we are currently interested in. If the story gets redeployed/restarted, we have the old estimate from the DB, but we don't want to base our new mean on the old values anyhow as they might have changed. If we want to log the values for anything more serious, as we agreed before, Postgres shouldn't be used anyhow.

asyncy/ServiceUsage.py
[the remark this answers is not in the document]
anukul: Yeah, a ring buffer / circular queue makes the most sense.

On every run, we need to dump data in the db so new pods can be deployed with the right limits.
Instead of fetching and dumping the complete array, we can store the current index of the circular queue and update the value at that index on every run.

I can go ahead and do that if it makes a significant performance difference, considering the size of the arrays.

asyncy/Apps.py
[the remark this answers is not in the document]
wilzbach: BTW to explain this a bit more: the problem of a just a separate thread is that if it crashes, there's no retry logic, diagnostic or health checking.

Depends a bit on what we will need in the future. One option could be to use a job queue (e.g. [celery](https://github.com/celery/celery) or [rq](https://github.com/rq/rq). A job queue makes sense if we want to make deploys truly asynchronous and distributed + plus we get a lot of features out of the box (e.g. retries, diagnostics, and auto-scheduling of these jobs on idle workers).
Other options are:
- wrap this as a separate container, add it to the same pod and start with the app. then kubernetes should take care of health-checks, retries and diagnostic for us (though this comes with quite some overhead)
- use a micro-scheduler (e.g. `cron` or [`aiojobs`)[https://github.com/aio-libs/aiojobs]))

BTW we already use Prometheus for the application logs, so why not store the usage data there as well if we plan to do it anyhow?

asyncy/db/Database.py
anukul: Cool, so it looks like we have a bunch of performance optimizations - parallel recording, circular queues, bulk updates.

asyncy/Apps.py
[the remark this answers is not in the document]
wilzbach: Opened an issue for moving away from spawning simple threads for services - https://github.com/storyscript/platform-engine/issues/356

chrisstpierre:
@wilzbach my worry about this issue is that we are essentially taking random data points and hoping they are an accurate description of resource use. We are going to have a lot of services that are mostly unused, with occasional spikes in resource use, & we need enough resources supplied to be able to manage those spikes. IE: ImageMagick might be used a few times per day, but when it is used it requires a ton of resources, and our data collection this way doesn't help us prepare for that really. Without enough resources ImageMagick will just crash. @anukul and I had a discussion about this as well.

RE: dynamically down/up-size pods - We definitely want to do this - but hold on because it's not that easy. Of course we could scale up when use increases, but that might not actually help anything. What we would really need for it to be useful is a machine learning service hooked up to our Prometheus data. Predictive Scaling: [aws predictive scaling](https://aws.amazon.com/blogs/aws/new-predictive-scaling-for-ec2-powered-by-machine-learning/) [ai kubernetes autoscaling](https://docs.magalix.com/docs/how-magalix-ai-powers-applications). If we don't do that then we are behind the curve. 

I would also suggest giving advanced users the ability to choose when their services scale. I'm not sure how it would look, but maybe in the story?
?
```
when http server listen method:'post' path:'/upload' as req
    dest = '/uploads/{awesome id}.png'
    image = req.files['image']
    image_bytes = image['size']

    needed_memory = round (image_bytes/1048576)*2
    needed_cpu = round needed_memory/3

    scale imagemagick mem:"{needed_memory}M" cpu:"{needed_cpu}m"
    pdf= imagemagik convert image:image['binary'] to:pdf width:2000 height:5000

    aws/s3 put content:img :dest
    req write content:{location:dest}
    req finish
```

Because for this type of scenario no predictive scaling can help. So what else could we do besides this?
We couldn't let the service scale itself either, because the service doesn't know the business case. If we need this image converted FAST then we need to give more resources. If we are okay with slower processing then we can choose something smaller to save money.

And then I suppose the service should go back to its default size after the line of last occurrence is finished executing. 

Ok this is a long comment I will make a new issue for this.

judepereira:
@wilzbach dynamic scale up and down is planned, but later. The problem we face today is that the `redis` service and many others are just python wrappers. They consume just 20MB, but we still allocate 200 MB for it.

judepereira:
@anukul what's the status of this? Is this in sync with the database?

judepereira:
@anukul Percentile is simple to implement, given a N values. I'd like to see that change.

asyncy/Apps.py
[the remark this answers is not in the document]
judepereira: @wilzbach to close the loop here, we do not use prometheus yet. I like the simple approach of a single thread, as long as it's ensured that it runs regardless of exceptions. Using celery is an overkill.

asyncy/utils/Stats.py
```diff
@@ -0,0 +1,16 @@
+import math
+
+
+def percentile(data: list, percent: float, sorted=False):
```
wilzbach: docstring

wilzbach: Also why do we have to reinvent the wheel here? What was wrong with numpy?

asyncy/ServiceUsage.py
```diff
@@ -0,0 +1,168 @@
...
+                'cpu_units': np.percentile(cpu_units[tag], 95),
+                'memory_bytes': np.percentile(memory_bytes[tag], 95)
+            }
+            for tag in cpu_units.keys()
+        ]
+
+    @classmethod
+    async def record_service_usage(cls, config: Config, logger: Logger):
+        from .Service import Service
+        while not Service.shutting_down:
```
judepereira: @anukul any exceptions will lead to this task dying permanently. You need to wrap this in a try/catch block, catching all exceptions, logging them, and sleeping for 5 seconds (`await asyncio.sleep(5)`).

asyncy/Kubernetes.py
```diff
@@ -596,8 +603,8 @@ def get_liveness_probe(cls, app, service: str):
                                 'image': image,
                                 'resources': {
                                     'limits': {
-                                        'memory': '200Mi',  # During beta.
-                                        # 'cpu': '500m',  # During beta.
+                                        'memory': limits['memory'],
+                                        'cpu': limits['cpu']
```
judepereira: @anukul Let's skip on cpu limits for now, since it's hard to judge (some take insane CPU to startup, and only use CPU when they're active).

judepereira:
Waiting on tests.

PR #349, opened 2019-06-18
Support for input output contract resolves #PL-185

rashmi43:


asyncy/processing/Services.py
```diff
@@ -427,7 +455,14 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
         if int(response.code / 100) == 2:
             content_type = response.headers.get('Content-Type')
             if content_type and 'application/json' in content_type:
-                return ujson.loads(response.body)
+                if cls.validate_output_properties(command_conf, response.body,
+                                                  story, line):
+                    return ujson.loads(response.body)
+                else:
+                    raise AsyncyError(message=f'Output contract violated! '
```
judepereira: What failed validation? What was missing? Which property was returned with an incorrect type? Let's make it easier to understand this. Note - the logs are the sole and only way of insight for a user into what's happening in the Storyscript Cloud.

asyncy/processing/Services.py
```diff
@@ -438,6 +473,34 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
 
+    @classmethod
+    def validate_output_properties(cls, command_conf: dict, body, story,
+                                   line):
+        """
+        Verify all properties are contained in the return body
+        """
+        output = command_conf.get('output', {})
+        flag = True
+        if output.get('properties') is not None:
```
judepereira: if `type` is object, and output has properties...

asyncy/processing/Services.py
```diff
@@ -438,6 +473,34 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+            props = output.get('properties')
+            key_list = props.keys()
+            if key_list:
```
judepereira: Too nested code, let's invert some if clauses so that the nesting level is not so high.

asyncy/processing/Services.py
```diff
@@ -438,6 +473,34 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+                for prop_key in key_list:
+                    if prop_key not in body:
+                        return False
+                    elif prop_key in body:
+                        prop_def = props[prop_key]
+                        prop_type = prop_def.get('type', None)
+                        if prop_type is not None:
+                            prop_value = body.get(prop_key)
+                            flag = cls.check_value_with_type(prop_value,
```
judepereira: Looks like there's no support for nested objects. Consider the following output as defined by the microservice.yaml:
```yaml
output:
  type: object
  contentType: application/json
  properties:
    a:
      type: string
    b:
      type: object
      properties:
        b1:
          type: string
```

asyncy/processing/Services.py
```diff
@@ -438,6 +473,34 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+                                                             prop_type)
+                            if flag is False:
```
judepereira: if not flag, return false.

asyncy/processing/Services.py
```diff
@@ -438,6 +473,34 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+        if output.get('properties') is not None:
+            props = output.get('properties')
+            key_list = props.keys()
+            if key_list:
+                for prop_key in key_list:
+                    if prop_key not in body:
+                        return False
+                    elif prop_key in body:
+                        prop_def = props[prop_key]
+                        prop_type = prop_def.get('type', None)
```
judepereira: .get will return None by default, no need to specify it explicitly.

asyncy/processing/Services.py
```diff
@@ -325,6 +325,34 @@ def raise_for_type_mismatch(cls, story, line, name, value, command_conf):
 
         raise ArgumentTypeMismatchError(name, t, story=story, line=line)
 
+    @classmethod
+    def check_value_with_type(cls, value, type):
```
judepereira: Might be worth moving to TypeResolver.py, and using `assert_type`. You can update `assert_type` to include more details in the exception that it raises.

asyncy/processing/Services.py
rashmi43: Removed high level of nesting in code

asyncy/processing/Services.py
rashmi43: incorporated

asyncy/processing/Services.py
rashmi43: added a proper error message

asyncy/processing/Services.py
rashmi43: I cannot use assert_type because I will be getting the data types as string and not keywords. Added my own error class.

asyncy/processing/Services.py
rashmi43: incorporated, changed the logic to not use flags.

asyncy/processing/Services.py
rashmi43: added support for nested properties

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+                                   line):
+        """
+        Verify all properties are contained in the return body
+        """
+        output = command_conf.get('output', {})
+        body_json = {}
+        if type(body) == str:
+            body_json = ujson.loads(body)
+        elif type(body) == dict:
+            body_json = body
```
wilzbach: When will an HTTP body ever be a `dict`?

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+            except TypeValueAssertionError as error:
+                raise AsyncyError(message=f'Output contract violated! '
+                                  f'Found value: {error.value}; '
+                                  f'Mismatched type: {error.type}',
+                                  story=story, line=line)
+        else:
+            return body_json
+
+    @classmethod
+    def recurse_nested(cls, type, val, body):
```
wilzbach: Should be a in a separate class.

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+        body_json = {}
+        if type(body) == str:
+            body_json = ujson.loads(body)
+        elif type(body) == dict:
+            body_json = body
+        if output.get('properties') is not None and output.get(
+                'type') == 'object':
+            try:
+                cls.recurse_nested('object', output, body_json)
+                return body_json
```
wilzbach: The verification doesn't need to return anything as by definition it only verifies.

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+            except TypeValueAssertionError as error:
+                raise AsyncyError(message=f'Output contract violated! '
```
wilzbach: Shouldn't this include more like e.g. the service name, action and the whole payload?

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+                                  f'Found value: {error.value}; '
+                                  f'Mismatched type: {error.type}',
+                                  story=story, line=line)
+        else:
+            return body_json
```
wilzbach: Not needed

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+
+    @classmethod
+    def recurse_nested(cls, type, val, body):
+        if type == 'object' and val.get('properties') is not None:
+            val_props = val.get('properties')
+            val_keys = val_props.keys()
+            for key in val_keys:
```
wilzbach: Could be one line instead of five:

```suggestion
            for key, key_def in val['properties'].item():
```

asyncy/utils/TypeResolver.py
```diff
@@ -52,6 +64,35 @@ def assert_type(expected, item):
...
 
+    @staticmethod
+    def check_value_with_type(value, type):
+        """
+        Validates for types listed on
+        https://microservice.guide/schema/actions/#arguments.
+
+        Supported types: int, float, string, list, map, boolean, or any
+        """
+        print(f'in check value is {value} and type is {type}')
```
wilzbach: Remove.

asyncy/utils/TypeResolver.py
```diff
@@ -52,6 +64,35 @@ def assert_type(expected, item):
                 return
         raise TypeAssertionError()
 
+    @staticmethod
+    def check_value_with_type(value, type):
```
wilzbach: should have `omg` in its name and imho be in an OMG/ServiceContract class as it doesn't have much to do with the Storyscript type resolver.

asyncy/utils/TypeResolver.py
```diff
@@ -52,6 +64,35 @@ def assert_type(expected, item):
...
+        elif type == 'int' and isinstance(value, int):
+            return
+        elif type == 'float' and isinstance(value, float):
+            return
+        elif type == 'list' and isinstance(value, list):
+            return
+        elif type == 'map' and isinstance(value, dict):
+            return
+        elif type == 'boolean' and isinstance(value, bool):
+            return
```
wilzbach: what about regex or time?

asyncy/utils/TypeResolver.py
```diff
@@ -52,6 +64,35 @@ def assert_type(expected, item):
...
+        elif type == 'any' or type == 'object':
+            return
+        else:
```
wilzbach: no else necessary here.

asyncy/utils/TypeResolver.py
```diff
@@ -52,6 +64,35 @@ def assert_type(expected, item):
...
+            raise TypeValueAssertionError(value, type)
```
wilzbach: should be named to what it actually does, e.g. `OMGTypeVerificationError`

tests/unit/processing/Services.py
```diff
@@ -431,6 +431,128 @@ def test_parse_output_invalid_type(story):
...
+                    }
+                }
+            }
+        }
+    }
+    body = '{"from": "storyscript", "msg_id": 123, "chat": "my chat"}'
+    body1 = '{"from": "storyscript", "msg_id": 123, "chat": "chat", \
+            "cid": 111, "cmsg":"hello"}'
+    expected_output = ujson.loads(body1)
+    print(f'expected_output is {expected_output}')
```
wilzbach: remove

tests/unit/processing/Services.py
```diff
@@ -431,6 +431,128 @@ def test_parse_output_invalid_type(story):
...
+
+    assert Services.validate_output_properties(
+        command_conf, body1, story, line) == expected_output
+    with pytest.raises(AsyncyError):
```
wilzbach: the error content isn't checked.

tests/unit/processing/Services.py
```diff
@@ -431,6 +431,128 @@ def test_parse_output_invalid_type(story):
...
+            }
+        }
+    }
+    body = '{"message_id": "msg_id_1"}'
+    body1 = '{"to": 123}'
+    body3 = {'status': True, 'message_id': 123}
+
+    with pytest.raises(AsyncyError):
+        Services.validate_output_properties(
+            command_conf, body, story, {})
```
wilzbach: every assertion should be in separate test. You can generate a test permutation to simplify this.

tests/unit/processing/Services.py
```diff
@@ -431,6 +431,128 @@ def test_parse_output_invalid_type(story):
...
+    # with pytest.raises(AsyncyError):
+    #    Services.validate_output_properties(
+    #        command_conf, body3, story, {})
+    assert Services.validate_output_properties(
+        command_conf, body3, story, {}) is body3
+
+
+@mark.parametrize('typ', ['int', 'float', 'string', 'list', 'map',
+                          'boolean', 'any', 'number'])
+@mark.parametrize('val', [1, 0.9, 'hello', [0, 1], {'a': 'b'}, True, False])
```
wilzbach: Use the tuple format, s.t. it's easier for humans to read and edit.

tests/unit/processing/Services.py
```diff
@@ -431,6 +431,128 @@ def test_parse_output_invalid_type(story):
...
+            'properties': {
+                'message_id': {
+                    'help': 'The message ID',
+                    'type': typ
+                }
+            }
+        }
+    }
+    body = {'message_id': val}
+    print(f'body expcted in test is {body}')
```
wilzbach: No poor man's debugging.

tests/unit/processing/Services.py
```diff
@@ -431,6 +431,128 @@ def test_parse_output_invalid_type(story):
...
+                              val, int)):
+        valid = True
+    elif typ == 'list' and isinstance(val, list):
+        valid = True
+    elif typ == 'map' and isinstance(val, dict):
+        valid = True
+    elif typ == 'boolean' and isinstance(val, bool):
+        valid = True
+    elif typ == 'any':
+        valid = True
```
wilzbach: Encode this in your test permutation, but don't just replicate your test code.

tests/unit/processing/Services.py
```diff
@@ -431,6 +431,128 @@ def test_parse_output_invalid_type(story):
...
+
+    if valid:
+        assert Services.validate_output_properties(
+            command_conf, body, story, {}) == body
```
wilzbach: The validation should always be called and checked.

asyncy/processing/Services.py
```diff
@@ -427,7 +428,8 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
         if int(response.code / 100) == 2:
             content_type = response.headers.get('Content-Type')
             if content_type and 'application/json' in content_type:
-                return ujson.loads(response.body)
+                return cls.validate_output_properties(
+                    command_conf, response.body, story, line)
```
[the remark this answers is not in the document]
wilzbach: Also `validate` isn't supposed to mutate, so no direct return.

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+    @classmethod
+    def recurse_nested(cls, type, val, body):
+        if type == 'object' and val.get('properties') is not None:
+            val_props = val.get('properties')
+            val_keys = val_props.keys()
+            for key in val_keys:
+                key_def = val_props[key]
+                prop_type = key_def['type']
+                if key not in body and prop_type != 'object':
+                    raise TypeValueAssertionError(key, prop_type)
```
wilzbach: This is a missing error whereas a few lines below it's a "ValueError". This means that the error message generated here won't make sense. Either use different exception classes or use more than value and type as fields.

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+                else:
+                    if prop_type == 'object' and key_def.get(
```
wilzbach:
```suggestion
                    if prop_type == 'object' and 'properties' in key_def
```

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
+                                    'properties') is not None:
+                        prop_value = key_def
+                    else:
+                        prop_value = body.get(key)
+                    if prop_type is None or prop_value is None:
+                        return TypeValueAssertionError(prop_value, prop_type)
+                    else:
+                        cls.recurse_nested(prop_type, prop_value, body)
```
wilzbach: The verification currently doesn't check whether the matching type is even an object. This will lead to weird errors when e.g. an `int` is returned but an object is expected.

asyncy/utils/TypeResolver.py
```diff
@@ -17,6 +17,18 @@ class TypeAssertionError(Exception):
     pass
 
 
+class TypeValueAssertionError(Exception):
```
wilzbach: Should be in Exceptions.py, no?

asyncy/processing/Services.py
rashmi43: So thats how I did it earlier but then we didn't know which property is missing or whose type is not matched. Because validate would only return a True or False.

tests/unit/processing/Services.py
rashmi43: Yes we can change it to complete random, currently it tests for positive cases. I can remove the if valid.

asyncy/processing/Services.py
rashmi43: sure will change this

asyncy/processing/Services.py
rashmi43: No it does that at the begin of the recurse_nested method. If first check if its a object , to see if its nested else if a simple value-type and calls check_value to match value with type.

asyncy/utils/TypeResolver.py
rashmi43: I noticed that all utils classes like TypeResolver have their exceptions defined in the same class. I have followed suit. There is a TypeAssertionError and its part of TypeResolver.

asyncy/processing/Services.py
```diff
@@ -438,6 +440,54 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
                               f'response body: {response_body}',
                               story=story, line=line)
 
+    @classmethod
+    def validate_output_properties(cls, command_conf: dict, body, story,
+                                   line):
+        """
+        Verify all properties are contained in the return body
```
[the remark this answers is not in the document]
rashmi43: Does this do nested properties? Doesn't look like it.

asyncy/processing/Services.py
rashmi43: No it won't be, but I noticed that if a string is passed we cannot see the value types everything is a string.

asyncy/processing/Services.py
rashmi43: Yes this is what I had done before but @judepereira wanted to know exactly which type and property failed, so  I had to throw the error from validate_output_properties method which knows exactly which property is missing and changed the logic to return the body from this method itself. And thats when i faced a lot of json parsing related issues and hence had to use dict for my testing.

So If i throw an error here we may not know exactly which prperty failed because this method will always return a True or False, unless I make it throw a TypeValueAssertionType error and error will have the type and property and I can report that here. Lets decide the strategy on this one? @judepereira and @wilzbach

asyncy/processing/Services.py
rashmi43: ok sure will change this pronto.

asyncy/utils/TypeResolver.py
rashmi43: yes sure :) missed it somehow when cleaning my code.

asyncy/utils/TypeResolver.py
rashmi43: ok will consider was looking for options to make it short and crisp.

asyncy/utils/TypeResolver.py
rashmi43: okay its a generic value to type matcher so didn't bring any OMG aspect to it. But I can do that the new method and exception can be in a new class.

tests/unit/processing/Services.py
rashmi43: ouch sure

tests/unit/processing/Services.py
rashmi43: ok will check the content as well.

tests/unit/processing/Services.py
rashmi43: what is a test permutation? you mean accept parameters to pass any inputs and permutate against those inputs?

tests/unit/processing/Services.py
rashmi43: okay sure i see this in a lot of other tests, thats where I got it from. I have tried to follow much of the coding style seen in other parts of the code.

tests/unit/processing/Services.py
rashmi43: yes i missed these, my bad

tests/unit/processing/Services.py
rashmi43: Please clarify test permutation for me? I think I understand but not completely sure.

asyncy/processing/Services.py
judepereira: Throw an exception with your message inside it, instead of trying to return what was invalid.

judepereira: Also, your recurse_method approach is a little weird. Can you make it like a normal recursive function?

asyncy/processing/Services.py
judepereira: Isn't the body always in string? If not string, it could only potentially be a `bytes` object. Never a dict.

wilzbach: No, then you must have done sth. else wrong. Loading a JSON will give you more than a dict. As this code isn't used by the production, it should be removed.

asyncy/processing/Services.py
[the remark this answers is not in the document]
judepereira: That's just for reference, such as enum value validation.

asyncy/processing/Services.py
wilzbach: You still don't know what's missing by returning the original JSON object. You do that by throwing exceptions. The point here was that there's no need to return the original JSON object as you're not supposed to touch it anyhow and if there's a validation error, you'll raise a validation error anyhow.

asyncy/utils/TypeResolver.py
judepereira: Oh, move them to Exceptions.py.

asyncy/processing/Services.py
wilzbach: > and hence had to use dict for my testing.

as mentioned earlier, you must have done sth. wrong (or encountered a bug somewhere else), but you should never need to do that.

> So If i throw an error here we may not know exactly which property failed

I don't think you understood me. Throwing an error in the validation process is fine, but my point was that this error should print more than just output contract violated as e.g. the user doesn't even know for which service and which call the validation failed.

asyncy/utils/TypeResolver.py
```diff
@@ -17,6 +17,18 @@ class TypeAssertionError(Exception):
...
+    """
+    Raised when value does not match type.
+    """
+    def __init__(self, value, type):
+        self.value = value
+        self.type = type
+
+    def __str__(self):
+        return(repr(self.value))
```
wilzbach:
```suggestion
        return str(self.value)
```

Though we should print a lot more here.

asyncy/utils/TypeResolver.py
wilzbach: > okay its a generic value to type matcher so didn't bring any OMG aspect to it.

This is only about the OMG specification. Even your first line of the description says so. It's not generic.

asyncy/utils/TypeResolver.py
wilzbach: FWIW I'm not sure whether OMG supports them yet, but it should soon.

asyncy/utils/TypeResolver.py
```diff
@@ -52,6 +64,35 @@ def assert_type(expected, item):
...
+    def check_value_with_type(value, type):
+        """
+        Validates for types listed on
+        https://microservice.guide/schema/actions/#arguments.
+
+        Supported types: int, float, string, list, map, boolean, or any
+        """
+        print(f'in check value is {value} and type is {type}')
+        if type == 'string' and isinstance(value, str):
+            return
```
wilzbach: this code could be sth. like: 

```py
if type == 'string':
   check_types(value, str)
```

This reads better, improves the coverage results and gets rid of the weird `return`

tests/unit/processing/Services.py
wilzbach:
```py
@mark.parametrize('body', [
   '{"message_id": "msg_id_1"}'
...
])
```

tests/unit/processing/Services.py
wilzbach: > a lot of other tests, thats where I got it from

This doesn't mean it's a good style. 

```py
@mark.parametrize('typ,val', [
  ('int', 1),
  ('float', 0.9),
])
```

Doesn't this read better?

tests/unit/processing/Services.py
wilzbach: If you have any test which is supposed to fail, either add this as additional `mark.parametrize` parameter or add a new test (i.e. a new method with all options that are supposed to fail).

tests/unit/processing/Services.py
rashmi43: yes it does

tests/unit/processing/Services.py
rashmi43: okay

asyncy/utils/TypeResolver.py
rashmi43: in my earlier commit I had done it that way, anyways I have changed back to that. I will only validate and not touch the json.

asyncy/Exceptions.py
```diff
@@ -120,6 +120,40 @@ def __init__(self, story=None, line=None, service=None, action=None):
...
+        super().__init__(
+            f'The property value {value}'
+            f'should be of type {type}.'
+            f'It does not adhere to the microservice '
+            f'service contract.')
+        self.value = value
+        self.type = type
+
+    def __str__(self):
+        return(str(self.value))
```
wilzbach:
```suggestion
        return repr(self.value)
```

Also, as mentioned in earlier review we shouldn't even overwrite `__str__` here. What's the reason?

asyncy/Exceptions.py
```diff
@@ -120,6 +120,40 @@ def __init__(self, story=None, line=None, service=None, action=None):
...
 
+class OmgPropertyValueMismatchError(AsyncyError):
+    """
+    Raised when value does not match type.
+    """
+    def __init__(self, value, type):
+        super().__init__(
+            f'The property value {value}'
+            f'should be of type {type}.'
+            f'It does not adhere to the microservice '
```
wilzbach:
```suggestion
            'It does not adhere to the microservice '
```

asyncy/Exceptions.py
```diff
@@ -120,6 +120,40 @@ def __init__(self, story=None, line=None, service=None, action=None):
...
+            f'service contract.')
```
wilzbach:
```suggestion
            'service contract.')
```

asyncy/Exceptions.py
```diff
@@ -120,6 +120,40 @@ def __init__(self, story=None, line=None, service=None, action=None):
...
+
+class OmgPropertyTypeMissingError(AsyncyError):
+    """
+    Raised when value does not match type.
+    """
+    def __init__(self, value, type):
+        super().__init__(
+            f'The property value {value} '
+            f'does not have a type {type} specifed.'
+            f'Please check microservice.yml '
```
wilzbach:
```suggestion
            'Please check microservice.yml '
```

asyncy/Exceptions.py
```diff
@@ -120,6 +120,40 @@ def __init__(self, story=None, line=None, service=None, action=None):
...
+            f'and specify type for each value')
```
wilzbach:
```suggestion
            'and specify type for each value')
```

asyncy/Exceptions.py
```diff
@@ -120,6 +120,40 @@ def __init__(self, story=None, line=None, service=None, action=None):
...
+        self.value = value
+        self.type = type
+
+    def __str__(self):
+        return(str(self.value))
```
wilzbach: ditto

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+from ..Exceptions import AsyncyError, OmgPropertyTypeMissingError, \
+    OmgPropertyValueMismatchError
+
+
+class ServiceContract:
+
+    @classmethod
+    def check_value_with_type(cls, typ, value):
```
wilzbach: better `type_` instead of having a typo.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Validates for types listed on
```
wilzbach:
```suggestion
        Ensures that the value matches the expected type as listed on
```

asyncy/omg/ServiceContract.py
wilzbach: Also, I presume `ensure_value_has_type` or `enforce_value_of_type`  would be more descriptive names.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+from ..Exceptions import AsyncyError, OmgPropertyTypeMissingError, \
+    OmgPropertyValueMismatchError
+
+
+class ServiceContract:
```
wilzbach: docs

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+        if typ == 'boolean':
+            cls.check_types(bool, value)
+        if typ == 'map':
+            cls.check_types(dict, value)
+        if typ == 'float':
+            cls.check_types(float, value)
+        if typ == 'list':
+            cls.check_types(list, value)
+        if typ == 'any':
+            return
```
wilzbach: uncovered. needs an exception here too.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+    def check_value_with_type(cls, typ, value):
+        """
+        Validates for types listed on
+        https://microservice.guide/schema/actions/#arguments.
+
+        Supported types: int, float, string, list, map, boolean, or any
+        """
+        if typ is None:
+            raise OmgPropertyTypeMissingError(typ)
+        if typ == 'string':
```
wilzbach:
```suggestion
        elif typ == 'string':
```

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+                                  story=story, line=line)
+            elif props.get(key).get('type') == 'object':
+                cls.validate_output_properties(
+                    props.get(key), body.get(key), story, line)
+            else:
+                cls.check_value_with_type(
+                    props.get(key).get('type'), body[key])
+
+    @staticmethod
+    def check_types(typ, val):
```
wilzbach: More descriptive name and description, e.g. `enforce_value_of_type`

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+            return
+
+    @classmethod
+    def validate_output_properties(cls, output: dict, body, story,
+                                   line):
+        """
+        Verify all properties are contained in the return body
+        """
+        # if body is None or {}:
+        #    return
```
wilzbach: Remove?

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+        if output is None or {}:
```
wilzbach: If I understand correctly, it's also an error if the user specified no output and the service returned output.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+            return
+        props = output.get('properties')
+        for key in props:
```
wilzbach:
```suggestion
        for key, value in props.items():
```

You save a lot of `props.get(key)` then.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+            if key not in body:
+                raise AsyncyError(message=f'Output contract violated! '
+                                  f'Property key not found: {key}; '
+                                  f'Expected key is missing in return ',
+                                  story=story, line=line)
+            elif props.get(key).get('type') == 'object':
+                cls.validate_output_properties(
+                    props.get(key), body.get(key), story, line)
```
wilzbach:
```suggestion
                    value, body[key], story, line)
```

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+        if output is None or {}:
+            return
+        props = output.get('properties')
+        for key in props:
+            if key not in body:
+                raise AsyncyError(message=f'Output contract violated! '
+                                  f'Property key not found: {key}; '
+                                  f'Expected key is missing in return ',
+                                  story=story, line=line)
+            elif props.get(key).get('type') == 'object':
```
wilzbach:
```suggestion
            elif value['type'] == 'object':
```

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+        # if body is None or {}:
+        #    return
+        if output is None or {}:
+            return
+        props = output.get('properties')
+        for key in props:
+            if key not in body:
+                raise AsyncyError(message=f'Output contract violated! '
+                                  f'Property key not found: {key}; '
+                                  f'Expected key is missing in return ',
```
wilzbach: As mentioned earlier, this still needs the name of the service contract and its returned body as otherwise this will be hard to debug.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,65 @@
...
+        if typ == 'int':
+            cls.check_types(int, value)
+        if typ == 'boolean':
+            cls.check_types(bool, value)
+        if typ == 'map':
+            cls.check_types(dict, value)
+        if typ == 'float':
+            cls.check_types(float, value)
+        if typ == 'list':
+            cls.check_types(list, value)
```
wilzbach: what about other types like e.g. `object`?

asyncy/processing/Services.py
```diff
@@ -424,10 +426,14 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
             3, story.logger, url, client, kwargs)
 
         story.logger.debug(f'HTTP response code is {response.code}')
+        output = command_conf.get('output')
```
wilzbach: keep the scope as minimal as possible -> move to the scope where it's actually needed

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,132 @@
+# -*- coding: utf-8 -*-
+import json
+import uuid
+from unittest.mock import MagicMock, Mock
+
+from asyncy.Exceptions import AsyncyError, OmgPropertyValueMismatchError
+from asyncy.omg.ServiceContract import ServiceContract
+
+import pytest
+from pytest import mark
```
wilzbach: why not:

```suggestion
from pytest import mark, raises
```

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,132 @@
...
+    body1 = '{"from": "storyscript", "msg_id": 123, "chat": "chat", \
+             "id": {"cid": 111, "cmsg": "hello"}}'
+    message = 'Output contract violated! Property key not found: id; \
+        Expected key is missing in return'
+
+    ServiceContract.validate_output_properties(
+        command_conf.get('output'), json.loads(body1), story, line)
+    with pytest.raises(AsyncyError):
+        ServiceContract.validate_output_properties(
+            command_conf.get('output'), json.loads(body), story, {}) == message
```
wilzbach: this is a different test and thus needs to be separate. Also it does nothing as there's no assert

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+        }
+    }
+    message = f'The property value message_id should be of type int.It does \
+        not adhere to the microservice service contract.'
+
+    with pytest.raises(OmgPropertyValueMismatchError):
+        ServiceContract.validate_output_properties(
+            command_conf.get('output'), body, story, {}) == message
```
wilzbach: this doesn't actually validate the actual exception message (no `assert`!). This needs to be done outside of the `with` scope.

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+        }
+    }
+    message = 'Output contract violated! Property key not found: message_id; \
+        Expected key is missing in return'
+
+    with pytest.raises(AsyncyError):
+        ServiceContract.validate_output_properties(
+            command_conf.get('output'), body, story, {}) == message
```
wilzbach: ditto

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,132 @@
...
+                    'help': 'The message ID',
+                    'type': 'int'
+                },
+                'status': {
+                    'type': 'boolean',
+                    'help': 'status'
+                }
+            }
+        }
+    }
```
wilzbach: As you copy this a few times, I recommend putting it outside of the functions to be DRY.

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@mark.parametrize('typ,val', [('int', 1.9), ('float', 1), ('string', 1),
+                              ('list', 'hello'), ('map', [0, 1]),
+                              ('boolean', 'True')])
+def test_fail_check_value_each_type(typ, val, story):
+    msg = f'The property value {val} should be of type {typ}.It does \
+        not adhere to the microservice service contract.'
+    with pytest.raises(OmgPropertyValueMismatchError):
+            ServiceContract.check_value_with_type(typ, val) == msg
```
wilzbach: ditto there's no assert and this doesn't check the exception message properly.

tests/unit/omg/ServiceContract.py
rashmi43: my validate_output_properties is not returning anything earlier it was returning True/False so I don't know what to assert :( but i will see what i can do.

tests/unit/omg/ServiceContract.py
rashmi43: yes I was trying to achieve that. I will put it outside and check.

asyncy/omg/ServiceContract.py
rashmi43: yes i was just changing that, all my exceptions should have Omg. i am making a new asyncy/OmgExceptions.py to include the 3 exceptions I have created.

asyncy/omg/ServiceContract.py
rashmi43: if type is any we need not check anything right? the value can be anything?

tests/unit/omg/ServiceContract.py
wilzbach:
```py
    with pytest.raises(OmgPropertyValueMismatchError) as e:   
        ... 
    assert e.value.message == '..' 
```

tests/unit/omg/ServiceContract.py
wilzbach:
```py
    with pytest.raises(AsyncyError) as e:   
        ... 
    assert e.value.message == '..' 
```

asyncy/omg/ServiceContract.py
wilzbach: Yes, but what if type != `any`. you don't check for that.

asyncy/omg/ServiceContract.py
rashmi43: okay

rashmi43: also the key length has to match exactly.

asyncy/omg/ServiceContract.py
rashmi43: done

asyncy/omg/ServiceContract.py
rashmi43: it won't come here if type is object, if type is object it will check if it has properties recursively.

asyncy/processing/Services.py
rashmi43: yes will do

tests/unit/omg/ServiceContract.py
rashmi43: i did it with match=message

asyncy/omg/ServiceContract.py
rashmi43: can you explain this: type != any
i have a check for type not specified.

wilzbach: What if `type` is not `any`? This isn't covered here. You could replace the `if` with an `assert`

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,74 @@
...
+        if _type == 'any':
+            return
+
+    @classmethod
+    def validate_output_properties(cls, output: dict, body, story,
+                                   line):
+        """
+        Verify all properties are contained in the return body
+        """
+        if (not body and output) or (not output and body):
```
wilzbach: This is an XOR and can be written more concisely:

```suggestion
        if bool(body) ^ bool(output)
```

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,74 @@
...
+            raise OmgMismatchedPropertyLengthError(len(output), len(body))
+        props = output.get('properties')
+        if len(props) != len(body):
+            raise OmgMismatchedPropertyLengthError(len(props), len(body))
```
wilzbach: How will the user know which properties were missing?

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,313 @@
...
+                                'cmsg_id': {
+                                    'type': 'int'
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
wilzbach: This is duplicated

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,313 @@
...
+    body = '{"from": "storyscript", "msg_id": 123, "chat": "chat", \
+             "id": {"cid": 111, "cmsg": {"name": "ss", "cmsg_id": 1}}}'
```
wilzbach: Why not use `parametrize` for this and the body above?

tests/unit/omg/ServiceContract.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+@mark.parametrize('typ,val', [('int', 1.9), ('float', 1), ('string', 1),
+                              ('list', 'hello'), ('map', [0, 1]),
+                              ('boolean', 'True')])
+def test_fail_check_value_each_type(typ, val, story):
+    msg = f'The property value {val} should be of type {typ}.It does \
+        not adhere to the microservice service contract.'
+    with pytest.raises(OmgPropertyValueMismatchError):
+            ServiceContract.ensure_value_of_type(typ, val) == msg
```
wilzbach: This isn't an `assert`

tests/unit/omg/ServiceContract.py
rashmi43: done

tests/unit/omg/ServiceContract.py
rashmi43: done using match=msg had missed this one function.

asyncy/omg/ServiceContract.py
rashmi43: added a log statement and specified in error message as well.

asyncy/omg/ServiceContract.py
rashmi43: done

tests/unit/omg/ServiceContract.py
rashmi43: changed it

tests/unit/omg/ServiceContract.py
rashmi43: done

tests/unit/omg/ServiceContract.py
rashmi43: done

asyncy/omg/OmgExceptions.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+from ..Exceptions import AsyncyError
+
+
+class OmgPropertyTypeNotFoundError(AsyncyError):
+    """
+    Raised when an unknown type is specified which is not
+    recognised as per the OMG specification.
+    """
+    def __init__(self, _type):
```
wilzbach:
```suggestion
    def __init__(self, type_):
```

Followin PEP8: https://www.python.org/dev/peps/pep-0008/#id36

> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,78 @@
...
+        elif _type == 'float':
+            cls.ensure_type(float, value, _type)
+        elif _type == 'list':
+            cls.ensure_type(list, value, _type)
+        elif _type == 'number':
+            cls.ensure_type_list([int, float], value, _type)
+        elif _type == 'any':
+            return
+        else:
+            raise OmgPropertyTypeNotFoundError(_type)
```
wilzbach: This isn't a user exception. This is a platform failure. Hence, this should be an internal assert as we know what types are supported by OMG.

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,78 @@
...
+    def ensure_type(typ, val, _type):
+        """
+        Check if value belongs to the type specified.
+        """
+        if isinstance(val, typ):
+            return
+        raise OmgPropertyValueMismatchError(_type, val)
+
+    @staticmethod
+    def ensure_type_list(_list, val, _type):
```
wilzbach:
```suggestion
    def ensure_type_list(list_, val, type_):
```

Follow PEP8

asyncy/omg/ServiceContract.py
```diff
@@ -0,0 +1,78 @@
...
+                                   line):
+        """
+        Verify all properties are contained in the return body
+        """
+        if bool(body) ^ bool(output):
+            raise OmgMismatchedPropertyLengthError(output, body)
+        props = output.get('properties')
+        if len(props) != len(body):
+            story.logger.error(f'The property items {props.keys()} do not '
+                               f'match the body item length {body.keys()}')
```
wilzbach: This won't print the entire body as the object might be nested.

asyncy/processing/Services.py
```diff
@@ -427,7 +429,11 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
         if int(response.code / 100) == 2:
             content_type = response.headers.get('Content-Type')
             if content_type and 'application/json' in content_type:
-                return ujson.loads(response.body)
+                output = command_conf.get('output')
+                body = ujson.loads(response.body)
+                ServiceContract.validate_output_properties(
```
wilzbach: So what I was going for the last few times was to check for all ContractViolations here, catch them and add all the context to them.
Here you have all the info that's helpful to the user: what service, what method, the entire body etc.

rashmi43: I have that in validate_output_properties story and line i can add to all messages. When I was testing story printed story = <asyncy.Stories.Stories object at 0x7fce05b3cd30>.

asyncy/omg/ServiceContract.py
rashmi43: Is it a good idea to print the whole body will it ever have anything confidential? Thats why till now I have refrained from printing the whole body.

asyncy/omg/ServiceContract.py
rashmi43: i am doing a string compare... so I cannot replace if with assert. But i can do assert type in ['int','float','boolean','string','map','list'] and then let it throw AssertionError?

wilzbach: Of course you can:

```py
assert type_ == 'any'
```

asyncy/processing/Services.py
rashmi43: I can add a general try except around this call and throw AsyncyError?. I had done that in my first commit then I removed it. Please tell me how I should do that.

Lets decide on a way to fix final exception thing and what information to include in the errors, I have made the rest of the changes. Hopefully we can close after that. I will make that change tomorrow.

asyncy/omg/ServiceContract.py
rashmi43: Done

asyncy/processing/Services.py
rashmi43: I can't catch all exceptions and throw a new one, a bit confused here. Will think about it and come up with a solution. Do suggest a way if you know one.

asyncy/omg/OmgExceptions.py
rashmi43: done

asyncy/omg/ServiceContract.py
judepereira: Nothing confidential for now. Print away. Also, only people who have access to an app can view its logs.

rashmi43: I made that change and printed it in my last commit. Let me know if I can help with anything else.

judepereira:
Merged into PR #365.

PR #351, opened 2019-06-21
Add service health check

anukul:


asyncy/Kubernetes.py
```diff
@@ -477,6 +477,25 @@ def format_ports(cls, ports: {int}):
...
+        return {
+            'httpGet': {
+                'path': health_check['path'],
+                'port': health_check['port']
+            },
+            'initialDelaySeconds': 2,
+            'timeoutSeconds': 10,
+            'periodSeconds': 30,
+            'successThreshold': 1,
+            'failureThreshold': 1
```
anukul: FYI default failure threshold is 3

asyncy/Kubernetes.py
```diff
@@ -477,6 +477,25 @@ def format_ports(cls, ports: {int}):
                         f'Failed to pull image {container_status["image"]}'
                     )
 
+    @classmethod
+    def get_liveness_probe(cls, app, service: str):
```
wilzbach: doc

asyncy/Kubernetes.py
```diff
@@ -477,6 +477,25 @@ def format_ports(cls, ports: {int}):
...
+        omg = app.services[service][ServiceConstants.config]
+        health_check = omg.get('health', {}).get('http')
```
wilzbach: FWIW this is a bad style as it allocates a new dict object every time (slow).
The answers here a better:

https://stackoverflow.com/questions/25833613/python-safe-method-to-get-value-of-nested-dictionary

asyncy/Kubernetes.py
```diff
@@ -477,6 +477,25 @@ def format_ports(cls, ports: {int}):
...
+        if health_check is None:
+            return None
+        assert health_check['method'].upper() == 'GET'
```
wilzbach: No `.upper` needed here. See RFC 7230 Section 3.1.1

https://tools.ietf.org/html/rfc7230#section-3.1.1

asyncy/Kubernetes.py
```diff
@@ -583,6 +604,10 @@ def format_ports(cls, ports: {int}):
             }
         }
 
+        if liveness_probe is not None:
+            payload['spec']['template']['spec']['containers'][0][
+                'livenessProbe'] = liveness_probe
```
wilzbach: weird formatting

tests/unit/Kubernetes.py
```diff
@@ -894,6 +908,33 @@ def exc(a, timeout=0):
...
+            }
+        }
+    }
+    liveness_probe_first = Kubernetes.get_liveness_probe(app, 'first')
+    liveness_probe_second = Kubernetes.get_liveness_probe(app, 'second')
+    assert liveness_probe_first is None
+    assert liveness_probe_second['httpGet'] == {
+        'path': '/healthz',
+        'port': 8000
+    }
```
wilzbach: should be two separate tests or at least use `mark.parametrize`

asyncy/Kubernetes.py
anukul: I copied this from the accepted answer, and it's also used in a lot of other places in the code.
I think exception handling might be slower.
Do you have any suggestions?

asyncy/Kubernetes.py
anukul: Any alternatives?

wilzbach: Plenty, e.g.

````py
containers = payload['spec']['template']['spec']['containers']
containers[0]['livenessProbe'] = liveness_probe
```

asyncy/Kubernetes.py
wilzbach: > I think exception handling might be slower.

Yes.

> Do you have any suggestions?

This should be faster:

```py
def safeget(obj, *keys):
    for key in keys:
        obj = obj.get(key)
        if obj is None:
           return None
    return obj

...
safeget(obj, 'health', 'http')
```

anukul: Ah, this reminds me of `Dict.find`!

asyncy/Kubernetes.py
anukul: And another statement to assign the new `containers` value to `payload`.

So, do I change it?

wilzbach: > And another statement to assign the new containers

Python objects are by reference, it's already updated.
If you don't like this, at least do a standard formatting e.g.:

```py
 payload['spec']['template']['spec']['containers'][0] \
     ['livenessProbe'] = liveness_probe
```

anukul: Works, thanks!

asyncy/Kubernetes.py
```diff
@@ -470,13 +471,40 @@ def format_ports(cls, ports: {int}):
...
+        health_check = Dict.find(omg, 'health.http')
+        if health_check is None:
+            return None
+        assert health_check['method'] == 'get'
+        return {
+            'httpGet': {
+                'path': health_check['path'],
+                'port': health_check['port']
+            },
+            'initialDelaySeconds': 2,
```
judepereira:
```suggestion
            'initialDelaySeconds': 10,
```

asyncy/Kubernetes.py
```diff
@@ -470,13 +471,40 @@ def format_ports(cls, ports: {int}):
...
+            'timeoutSeconds': 10,
```
judepereira:
```suggestion
            'timeoutSeconds': 30,
```

asyncy/Kubernetes.py
```diff
@@ -470,13 +471,40 @@ def format_ports(cls, ports: {int}):
...
+            'periodSeconds': 30,
+            'successThreshold': 1,
+            'failureThreshold': 1
```
judepereira:
```suggestion
            'failureThreshold': 5
```

asyncy/Kubernetes.py
```diff
@@ -470,13 +471,40 @@ def format_ports(cls, ports: {int}):
...
+        If the liveness probe fails, the kubelet kills the Container,
+        and the Container is subjected to its restart policy.
+        If a Container does not provide a liveness probe,
+        the default state is Success.
+        """
+        omg = app.services[service][ServiceConstants.config]
+        health_check = Dict.find(omg, 'health.http')
+        if health_check is None:
+            return None
+        assert health_check['method'] == 'get'
```
judepereira:
```suggestion
        assert health_check.get('method', 'get') == 'get'
```

tests/unit/Kubernetes.py
```diff
@@ -894,6 +908,43 @@ def exc(a, timeout=0):
...
+                'port': 8000
+            }
+        }
+    },
+    'liveness_probe': {
+        'httpGet': {
+            'path': '/healthz',
+            'port': 8000
+        },
+        'initialDelaySeconds': 2,
```
judepereira:
```suggestion
        'initialDelaySeconds': 10,
```

tests/unit/Kubernetes.py
```diff
@@ -894,6 +908,43 @@ def exc(a, timeout=0):
...
+        'timeoutSeconds': 10,
+        'periodSeconds': 30,
+        'successThreshold': 1,
+        'failureThreshold': 1
```
judepereira:
```suggestion
        'failureThreshold': 5
```

tests/unit/Kubernetes.py
```diff
@@ -626,9 +626,22 @@ def test_new_ssl_context():
         })
     ]
 
+    liveness_probe = {
+        'httpGet': {
+            'path': '/healthz',
+            'port': 8000
+        },
+        'initialDelaySeconds': 2,
```
judepereira:
```suggestion
        'initialDelaySeconds': 10,
```

tests/unit/Kubernetes.py
```diff
@@ -626,9 +626,22 @@ def test_new_ssl_context():
...
+        'timeoutSeconds': 10,
```
judepereira:
```suggestion
        'timeoutSeconds': 30,
```

tests/unit/Kubernetes.py
```diff
@@ -626,9 +626,22 @@ def test_new_ssl_context():
...
+        'periodSeconds': 30,
+        'successThreshold': 1,
+        'failureThreshold': 1
```
judepereira:
```suggestion
        'failureThreshold': 5
```

tests/unit/Kubernetes.py
```diff
@@ -894,6 +908,43 @@ def exc(a, timeout=0):
...
+            }
+        }
+    },
+    'liveness_probe': {
+        'httpGet': {
+            'path': '/healthz',
+            'port': 8000
+        },
+        'initialDelaySeconds': 2,
+        'timeoutSeconds': 10,
```
judepereira:
```suggestion
        'timeoutSeconds': 30,
```

asyncy/App.py
```diff
@@ -90,8 +90,8 @@ def __init__(self, app_id: str, app_dns: str, version: int, config: Config,
                 message=f'Configuration for expose "{e.service_expose_name}" '
                 f'not found in service "{e.service}"')
 
-        target_path = conf.get('http', {}).get('path')
-        target_port = conf.get('http', {}).get('port')
+        target_path = Dict.find(conf, 'http.path')
```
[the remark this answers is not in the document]
anukul: @judepereira Apparently this is [slower](https://github.com/storyscript/platform-engine/pull/351#discussion_r296197199).

asyncy/Kubernetes.py
judepereira: No, Dict.find splits the string internally, which would still be ineffective. *keys is a better approach.

asyncy/App.py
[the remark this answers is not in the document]
judepereira: Dict.find splits a string, so it cannot be any better. If you're trying to optimise for this specifically:
1. There is no need as this is a micro optimisation - not a hot method
2. Even if this was a hot method, pick up optimisation in another PR (since there will be other use cases of Dict.find being slow in other hot methods)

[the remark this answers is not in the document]
anukul: I meant that the old code was slower, as mentioned in the linked comment.
Sorry about mixing it up in one PR though.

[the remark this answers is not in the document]
judepereira: Yes, but slow/fast doesn't matter here.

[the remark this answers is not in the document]
wilzbach: @anukul: I never said that Dict.find is faster. Just that allocating a lot of empty objects is a waste and bad style as it could easily be avoided.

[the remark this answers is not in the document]
anukul: Sorry for the misinterpretation

PR #352, opened 2019-06-21
Close DB connections

judepereira:
Since we're closing every database connection, we will eventually need to have a threaded connection pool.


wilzbach:
> Since we're closing every database connection, we will eventually need to have a threaded connection pool.

-> opened as a ticket, s.t. we don't forget this: https://github.com/storyscript/platform-engine/issues/353
(though it might be worth investigating whether we should switch to the async driver or a ORM before doing so)

judepereira:
> This is ugly. Can't we do with with conn.cursor() as cur?

You can, but nothing is stopping one from doing the following too:
```py
with new_conn() as conn:
    with conn.cursor() as cur:
         # too much indentation already
```

wilzbach:
>   # too much indentation already

1) we should use a connection pool anyhow
2) you can wrap both into one scope for now:

```py
with connect_cursor(app.config) as cur:
    ...
```

where `connect_cursor` is:

```py
class SimpleConnCursor:
    def __init__(self, config):
        self.config = config

    def __enter__(self):
        self.conn = cls.new_pg_conn(self.config)
        self.cur = conn.cursor(cursor_factory=RealDictCursor)
        return self.cur

   def __exit__(self, exc_type, exc_value, traceback):
       self.cur.close()
       self.conn.close()

...
def connect_cursor(config):
   return SimpleConnCursor(config)
```

Apart from making testing easier, because you only need to test the `SimpleConnCursor`, this will also make updating to a threaded connection pool easier as there's only method to touch.

judepereira:
@wilzbach Needed a whole ton of refactoring, took quite some time. Anyhow, review :)

asyncy/db/Database.py
```diff
@@ -0,0 +1,84 @@
...
+class Database:
+
+    @classmethod
+    def new_pg_conn(cls, config: Config):
+        conn = psycopg2.connect(config.POSTGRES)
+        return conn
+
+    @classmethod
+    def get_all_app_uuids_for_deployment(cls, config: Config):
+        with SimpleConnCursor(cls.new_pg_conn(config)) as db:
```
wilzbach: Why not do the following:

```py
    @classmethod
    def new_pg_conn(cls, config: Config):
        conn = psycopg2.connect(config.POSTGRES)
        return conn

    @classmethod
    def new_cursor_conn(cls, config: Config) -> SimpleConnCursor:
        return SimpleConnCursor(cls.new_pg_conn(config))

    @classmethod
    def get_all_app_uuids_for_deployment(cls, config: Config):
        with cls.new_cursor_conn(config) as db:
            ...
```

This should make testing easier as you only need to patch one method.

judepereira: Done.

tests/unit/Database.py
```diff
@@ -1,11 +1,10 @@
...
 from asyncy.enums.ReleaseState import ReleaseState
 
-from pytest import mark
 
-
-@mark.asyncio
-async def test_update_release_state(patch, magic, logger, config):
+def test_update_release_state(patch, magic, logger, config):
     conn = magic()
     patch.object(Database, 'new_pg_conn', return_value=conn)
```
wilzbach: If you mock just the context, you can do:

```py
db = magic()
patch.object(Database, 'new_cursor_conn', return_value=conn)
Database.new_cursor_conn.return_value.__enter__.return_value = db
```

judepereira: ?

tests/unit/db/SimpleConnCursor.py
```diff
@@ -0,0 +1,14 @@
...
+from asyncy.db.SimpleConnCursor import SimpleConnCursor
+
+from psycopg2.extras import RealDictCursor
+
+
+def test_simple(magic):
+    conn = magic()
+    with SimpleConnCursor(conn) as db:
+        assert db.conn == conn
+        assert db.cur == conn.cursor(cursor_factory=RealDictCursor)
```
wilzbach: magic's don't save their parameters, so this is equivalent to:

```py
assert db.cur == conn.cursor()
```

so you prepend this with a:

```py
conn.cursor.assert_called_with(cursor_factory=RealDictCursor)
```

judepereira: Ah, interesting. Thanks!

tests/unit/db/SimpleConnCursor.py
```diff
@@ -0,0 +1,14 @@
...
+from asyncy.db.SimpleConnCursor import SimpleConnCursor
+
+from psycopg2.extras import RealDictCursor
+
+
+def test_simple(magic):
+    conn = magic()
+    with SimpleConnCursor(conn) as db:
+        assert db.conn == conn
+        conn.cursor.assert_called_with(cursor_factory=RealDictCursor)
```
wilzbach: You still should do a an assertment on `db.cur`:

```py
assert db.cur == conn.cursor()
```

judepereira:
@anukul can you pull this branch locally, run the engine with a simple hello world story, and verify that it's working fine? My local env is broken, so I can't run a complete end to end test.

tests/unit/db/SimpleConnCursor.py
```diff
@@ -0,0 +1,14 @@
...
+
+    conn.cursor().close.assert_called_once()
+    conn.close.assert_called_once()
```
wilzbach: might be worthwhile to add a test with an exception to ensure the connections are closed there as well, e.g. sth. like:

```py
def test_close_on_error(magic):
    conn = magic()
    try:
        with SimpleConnCursor(conn) as db:
           raise Exception('test exception')
   finally:
        conn.cursor().close.assert_called_once()
        conn.close.assert_called_once()
```

tests/unit/Database.py
```diff
@@ -1,33 +1,112 @@
...
-from pytest import mark
+from pytest import fixture
 
 
-@mark.asyncio
-async def test_update_release_state(patch, magic, logger, config):
-    conn = magic()
-    patch.object(Database, 'new_pg_conn', return_value=conn)
+@fixture
+def database(magic, patch):
```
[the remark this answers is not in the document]
judepereira: Ah, this is okay.

tests/unit/db/SimpleConnCursor.py
judepereira: Done.

judepereira:
Thanks, looks good :)

judepereira:
@anukul awaiting your go ahead on your test results.

anukul:
Also works fine!

PR #362, opened 2019-06-28
Rebranding AsyncyError to StoryscriptError

TonyRice:

PR #363, opened 2019-06-30
Adding App Name, and Owner Email To Release/App Info

TonyRice:
This is required for the latest exception reporting updates. 

I've also merged https://github.com/storyscript/platform-engine/tree/err_rebrand into this branch to make things easier. Ideally, we should get pr #362 merged into master first.

### DIFF
--- a/asyncy/App.py
+++ b/asyncy/App.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 import asyncio
 import json
+import typing
 from collections import namedtuple
 
-import typing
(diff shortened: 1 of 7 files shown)


judepereira:
@TonyRice small change here - since the App constructor is now taking several arguments, can you create a namedtuple for this? Example: https://github.com/storyscript/platform-engine/blob/36258411145dfc16fb6807e950becdf711eab765/asyncy/entities/ContainerConfig.py

Since we'll want to add more and more int he future, this is probably something we should do whenever we touch code that accepts more than 5 arguments.

PR #364, opened 2019-07-01
Reporting

TonyRice:
This includes the following changes:

- Storyscript Asyncy Exception Rebranding/Refactor
- Exception Reporting with the following agents: Slack (Uses Webhooks), Sentry, and CleverTap
- Option for certain events to only be reported to CleverTap (clever_ident, and clever_event)
- User-enabled Exception/Error Reporting using Slack (REPORTING_SLACK_WEBHOOK within app configuration variable), without full system Stacktrace. Includes stack trace updates from @judepereira 
- User-enabled Reporting disabled by default (platform-engine USER_REPORTING_ENABLED option)
- Apps/Releases now include owner_email and app_name for debugging and Exception Reporting

Exception Reporting Slack Example:
Screenshot

User Defined Exception Reporting Slack Example:
Screenshot (1)

Sentry Reporting Example (In Email):
Screenshot (2)


TonyRice:
Note:

This PR has PR #362 and PR #363 merged into it for easier review and merge. Once those PRs are approved and merged, it'll be easier to break down these changes.

TonyRice:
Forgot to commit the tests inside this branch. Improving code coverage now.

asyncy/Apps.py
```diff
@@ -159,9 +178,19 @@ def make_logger_for_app(cls, config, app_id, version):
...
-        Sentry.init(sentry_dsn, release)
+        # initialize the engine's exception reporting system
+        ExceptionReporter.init({
+            'sentry_dsn': config.REPORTING_SENTRY_DSN,
+            'slack_webhook': config.REPORTING_SLACK_WEBHOOK,
+            'clevertap_config': {
+                'account': config.REPORTING_CLEVERTAP_ACCOUNT,
+                'pass': config.REPORTING_CLEVERTAP_PASS
+            },
+            'user_reporting': config.USER_REPORTING_ENABLED,
```
TonyRice: This option is used to disable the User App Exception Reporting. An env variable can be passed by using REPORTING_SLACK_WEBHOOK within their application. That will allow certain exceptions to be published to their webhooks.

asyncy/Apps.py
```diff
@@ -159,9 +178,19 @@ def make_logger_for_app(cls, config, app_id, version):
...
+            'user_reporting_stacktrace': config.USER_REPORTING_STACKTRACE
```
TonyRice: When this option is not false, a full stacktrace will be provided within the user reported exception. By default, it only publishes the storyscript error for the user. All other system exceptions reported to our comms will include a full exception.

asyncy/http_handlers/BaseHandler.py
```diff
@@ -22,16 +20,27 @@ def handle_story_exc(self, app_id, story_name, e):
...
+
+        app = Apps.get(app_id)
+
+        agent_options = {
+            'app_uuid': app_id,
+            'app_name': app.app_name,
+            'app_version': app.version,
+            'clever_ident': app.owner_email,
+            'clever_event': 'App Request Failure',
+            'allow_user_agents': True
```
TonyRice: When this is passed to the agent_options on a capture_exc, it tells the ExceptionReporter to allow exceptions to be published to user-defined agents, like the REPORTING_SLACK_WEBHOOK option

asyncy/Apps.py
```diff
@@ -52,9 +53,21 @@ def get_app_config(cls, raw):
...
         app_id = release.app_uuid
         stories = release.stories
-
-        logger = cls.make_logger_for_app(config, app_id, release.version)
-        logger.info(f'Deploying app {app_id}@{release.version}')
+        version = release.version
+
+        logger = cls.make_logger_for_app(config, app_id, version)
+
+        logger.log('app-deploying', app_id, version,
```
judepereira: Please no - don't use these labels. Finding where a log message was emitted from is a two step process using these labels, and one step without using labels.

asyncy/Logger.py
```diff
@@ -61,6 +62,9 @@ def format(self, record):
 
 class Logger:
     events = [
+        ('app-deploying', 'info', 'Deploying app {}@{}'),
```
judepereira: Please don't use labels. This is a poor approach. These existing labels are here because we haven't found the time to remove them.

asyncy/Logger.py
```diff
@@ -108,14 +113,49 @@ def adapt(self, app_id, version):
     def log(self, event, *args):
         self.frustum.log(event, *args)
 
+        try:
+            # Broadcast any desired events here using
+            # the reporting system.
+            from .reporting.Reporter import Reporter
```
judepereira: Why local import?

asyncy/Logger.py
```diff
@@ -108,14 +113,49 @@ def adapt(self, app_id, version):
...
+              reporting_agent_options: ReportingAgentOptions = None):
         getattr(self.frustum.logger, 'error')(message, exc_info=exc)
 
+        if (reporting_agent_options is not None and
+            reporting_agent_options.suppress_events is True) or \
+                self.reporting_enabled is False:
+            return
+
+        # local import
+        from .reporting.Reporter import Reporter
```
judepereira: Why a local import?

asyncy/reporting/ReportingAgent.py
```diff
@@ -0,0 +1,95 @@
...
+        from a traceback string.
+
+        :param traceback_str:
+        :return:
+        """
+        cwd = os.path.normpath(os.getcwd())
+        return re.sub(cwd, '', traceback_str)
+
+    @staticmethod
+    def format_tb_error(exc_info,
```
judepereira: Why not just call `str(exception)`?

asyncy/reporting/agents/CleverTapAgent.py
```diff
@@ -0,0 +1,128 @@
...
+                'method': 'POST',
+                'body': json.dumps({'d': [event]}),
+                'headers': {
+                    'X-CleverTap-Account-Id': self._account_id,
+                    'X-CleverTap-Passcode': self._account_pass,
+                    'Content-Type': 'application/json; charset=utf-8'
+                }
+            })
+
+    async def publish_exc(self, exc_info: BaseException,
```
judepereira: Duplicate function. Also, let's not over-engineer. Let's say that exceptions are not sent anywhere, but just used as an alerting mechanism that something happened. They can then use the logs to find the exception.

asyncy/reporting/agents/SentryAgent.py
```diff
@@ -0,0 +1,91 @@
...
+            story_line = exc_data['story_line']
+
+        if 'story_name' in exc_data:
+            story_name = exc_data['story_name']
+
+        full_stacktrace = True
+        suppress_stacktrace = False
+
+        if agent_config is not None:
+            if agent_config.get('full_stacktrace', True) is False:
```
judepereira: Send all stack trace info to Sentry. No need for special options such as full/suppress.

asyncy/reporting/agents/SentryAgent.py
```diff
@@ -0,0 +1,91 @@
...
+                'app_uuid': app_uuid,
+                'app_name': app_name,
+                'app_version': app_version,
+                'story_name': story_name,
+                'story_line': story_line
+            })
+            # we utilize captureMessage because captureException
+            # will not properly work 100% of the time
+            # unless this is always called within try/catch block
+            self._sentry_client.captureMessage(message=err_str)
```
judepereira: Have you seen the Sentry implications of using capture message vs capture exception? Sentry does some good work in formatting and providing context. They also allow grouping by similar exceptions, count occurrences, etc. Capture message might just be a plain message.

asyncy/reporting/agents/SlackAgent.py
```diff
@@ -0,0 +1,177 @@
+import json
```
judepereira: Save this agent for another PR, we'll start with Sentry and CleverTap.

asyncy/utils/HttpUtils.py
```diff
@@ -29,9 +29,11 @@ def read_response_body_quietly(response):
                 return res
             except HTTPError as e:
                 last_exception = e
-                logger.error(
-                    f'Failed to call {url}; attempt={attempts}; err={str(e)}'
-                )
+                if logger is not None:
```
judepereira: Why will logger be none? This should be avoided at all costs, since we want logging for every failed HTTP call to go somewhere (be it user logs, or engine logs).

tests/unit/Apps.py
```diff
@@ -387,11 +383,21 @@ def test_get_app_config(patch):
 @mark.parametrize('maintenance', [True, False])
 @mark.parametrize('always_pull_images', [True, False])
 @mark.parametrize('deleted', [True, False])
+@mark.parametrize(
+    'environment',
+    [
+        None,
+        {
+            'REPORTING_SLACK_WEBHOOK': 'http://example.com/'
```
judepereira: Ditch slack for now, save it for another PR as this has gotten large already.

tests/unit/Exceptions.py
```diff
@@ -51,6 +51,7 @@ def test_exception_trace(magic, patch, story, with_root):
     if root_message:
         root_message = f': {root_message}'
     assert str_version == f"""An exception has occurred:
+
```
judepereira: Why this new line? If it's purely subjective, don't do it in this PR.

tests/unit/Logger.py
```diff
@@ -50,69 +52,84 @@ def test_logger_init(logger, config):
...
+
+def test_logger_events_app_destroyed(logger):
+    message = 'Completed destroying app {}'
+    assert logger.events[2] == ('app-destroyed', 'info', message)
+
+
 def test_logger_events_container_start(logger):
     message = 'Container {} is executing'
-    assert logger.events[0] == ('container-start', 'info', message)
+    assert logger.events[3] == ('container-start', 'info', message)
```
judepereira: This is one other reason to **not** use logging labels. They were just put there earlier for unit tests, which are pointless to test (increases time and code for testing).

tests/unit/Logger.py
```diff
@@ -198,6 +215,34 @@ def test_logger_log(patch, logger):
     Frustum.log.assert_called_with('my-event')
 
 
+def test_logger_log_with_evt_reporting(patch, magic, logger):
```
judepereira: What is `evt_reporting`? I think you've merged two different concepts, and then have tried to adjust code to suit both, leading to over-engineering. Split "exception reporting", and "general alerts". Sending general alerts must be as simple as "Slack.send("some event occurred", "#alerts")". Sending events to CleverTap must be as simple as "CleverTap.push_event('App Deployed', {'State': 'Successful'})".

By definition, both are different, and force fitting them into each other will cause a lot of chaos in maintenance.

asyncy/utils/HttpUtils.py
TonyRice: Better safe than sorry to ensure that there's an actual logger passed along in this method.

tests/unit/Exceptions.py
TonyRice: Better formatting in the exception reporting

asyncy/utils/HttpUtils.py
wilzbach: Na, this should __crash__ the engine directly and loudly as it would be a big programming error in the engine. We shouldn't program so defensively.

tests/unit/Logger.py
TonyRice: At the end of the day, the "Reporting" system is as generic as possible, and the concept of the reporting agents are to simplify things a bit making it easy to update one agent, rather than update many lines of code throughout the entire platform. The goal is to actually have less maintenance, and the reporting system makes it easier to maintain a generic reporting system that can extend to different APIs.

asyncy/reporting/ReportingAgent.py
TonyRice: We don't get the full stack trace that way, just the message that's generally passed along. This will format an entire stacktrace, and it's root. If necessary, it'll also only format the generic message.

asyncy/Apps.py
```diff
@@ -121,20 +147,38 @@ def get_app_config(cls, raw):
...
+                         reporting_agent_options=ReportingAgentOptions(
+                             app_name=release.app_name,
+                             app_uuid=app_id,
+                             app_version=release.version,
+                             agent_config={
+                                 'clever_ident': release.owner_email,
+                                 'clever_event': 'App Deploy Failed'
+                             },
+                             allow_user_events=True
+                         ))
```
wilzbach: This is always the same. Can't we wrap it in a method?

tests/unit/Logger.py
judepereira: Yes, but we aren't adding agents every day, to make full use of the framework. Just two classes, three touch points in key code areas, would be far more easier to consume and maintain.

asyncy/reporting/Reporter.py
```diff
@@ -0,0 +1,502 @@
...
+                app_name = agent_options.app_name
+
+            if agent_options.app_uuid is not None:
+                app_uuid = agent_options.app_uuid
+
+            if agent_options.app_version is not None:
+                app_version = agent_options.app_version
+
+            if agent_options.agent_config is not None:
+                agent_config = agent_options.agent_config
```
wilzbach: Make a for loop out of this to reduce the boilerplate code.

asyncy/reporting/ReportingAgent.py
judepereira: This appears to work for all cases:
https://github.com/storyscript/platform-engine/blob/73a00c6d3e008b3a83a514e79bdc9cb1b8aa669d/asyncy/Logger.py#L34-L40

asyncy/reporting/Reporter.py
```diff
@@ -0,0 +1,502 @@
...
+            _evt_data['app_name'] = app_name
+
+        if app_uuid is not None:
+            _evt_data['app_uuid'] = app_uuid
+
+        if app_version is not None:
+            _evt_data['app_version'] = app_version
+
+        if story_name is not None:
+            _evt_data['story_name'] = story_name
```
wilzbach: Could be a for loop to reduce the boilerplate?

tests/unit/Exceptions.py
judepereira: Yes, that's subjective, so it's not required as such.

asyncy/reporting/Reporter.py
```diff
@@ -0,0 +1,502 @@
...
+                try:
+                    await agent_data.agent.publish_evt(
+                        evt_name=evt_name,
+                        evt_data=_evt_data,
+                        agent_config=agent_config
+                    )
+                except Exception as e:
+                    logger.error(
+                        message=f'Unhandled {agent_data.name} '
+                        f'reporting agent error: '
```
wilzbach:
```suggestion
                        'reporting agent error: '
```

asyncy/reporting/Reporter.py
```diff
@@ -0,0 +1,502 @@
...
+            if story.app.app_name is not None:
+                app_name = story.app.app_name
+
+            if story.app.app_id is not None:
+                app_uuid = story.app.app_id
+
+            if story.app.version is not None:
+                app_version = story.app.version
+
+            story_name = story.name
```
wilzbach: Too much boilerplate

asyncy/reporting/Reporter.py
```diff
@@ -0,0 +1,502 @@
...
+            exc_data['story_line'] = story_line
+
+        if app_name is not None:
+            exc_data['app_name'] = app_name
+
+        if app_uuid is not None:
+            exc_data['app_uuid'] = app_uuid
+
+        if app_version is not None:
+            exc_data['app_version'] = app_version
```
wilzbach: Too much boilerplate.

asyncy/reporting/agents/CleverTapAgent.py
```diff
@@ -0,0 +1,128 @@
...
+            evt_data['App name'] = exc_data['app_name']
+
+        if 'app_version' in exc_data:
+            evt_data['App version'] = exc_data['app_version']
+
+        if 'story_name' in exc_data:
+            evt_data['Story name'] = exc_data['story_name']
+
+        if 'story_line' in exc_data:
+            evt_data['Story line'] = exc_data['story_line']
```
wilzbach: Boilerplate.

asyncy/reporting/agents/SentryAgent.py
```diff
@@ -0,0 +1,91 @@
...
+            app_uuid = exc_data['app_uuid']
+
+        if 'app_version' in exc_data:
+            app_version = exc_data['app_version']
+
+        if 'story_line' in exc_data:
+            story_line = exc_data['story_line']
+
+        if 'story_name' in exc_data:
+            story_name = exc_data['story_name']
```
wilzbach: boilerplate

asyncy/reporting/agents/SlackAgent.py
```diff
@@ -0,0 +1,177 @@
...
+            app_uuid = f'*App UUID*: {evt_data["app_uuid"]}\n'
+
+        if 'app_version' in evt_data:
+            app_version = f'*App Version*: {evt_data["app_version"]}\n'
+
+        if 'story_name' in evt_data:
+            story_name = f'*Story Name*: {evt_data["story_name"]}\n'
+
+        if 'story_line' in evt_data:
+            story_line = f'*Story Line Number*: {evt_data["story_line"]}\n\n'
```
wilzbach: Boilerplate

asyncy/reporting/agents/SlackAgent.py
```diff
@@ -0,0 +1,177 @@
...
+        evt_str = ''
+        if evt_data['event_data'] is not None and \
+                len(evt_data['event_data'].items()) > 0:
+            evt_str = json.dumps(evt_data['event_data'])
+            evt_str = f'\n\n```{evt_str}```'
+
+        event_line = f'*Event*: {evt_name}{evt_str}'
+
+        err_msg = f'An event was triggered with ' \
+            f'the following information:\n\n' \
```
wilzbach:
```suggestion
            'the following information:\n\n' \
```

asyncy/reporting/agents/SlackAgent.py
```diff
@@ -0,0 +1,177 @@
...
+            app_uuid = f'*App UUID*: {exc_data["app_uuid"]}\n'
+
+        if 'app_version' in exc_data:
+            app_version = f'*App Version*: {exc_data["app_version"]}\n'
+
+        if 'story_name' in exc_data:
+            story_name = f'*Story Name*: {exc_data["story_name"]}\n'
+
+        if 'story_line' in exc_data:
+            story_line = f'*Story Line Number*: {exc_data["story_line"]}\n\n'
```
wilzbach: boilerplate

asyncy/Logger.py
TonyRice: The reporter uses the logger with reporting disabled. Don't want to cause an import loop

asyncy/utils/HttpUtils.py
TonyRice: I believe if we're aiming for a stable production-ready platform, coding defensively is exactly what we need to do. Especially if enterprise readiness is part of our roadmap.

asyncy/reporting/agents/SentryAgent.py
TonyRice: I have not, but the reason I went with captureMessage was due to a bug in the Raven client, and it wasn't reporting the exception properly. I can do my best to figure that out and move from there.

asyncy/utils/HttpUtils.py
judepereira: What happens in this particular scenario:
1. All logs are lost - lost executions, lost behaviour
2. 100% silent. Nobody is alerted. At least, if the logger throws an exception, we're notified of it instantly (if stories.storyscriptapp.com goes down, I react to it instantly)

TonyRice: Also, in the reporting system, I suppress log messages from http callbacks because I don't think the engine needs to crash if a reporting doesn't receive an event.

judepereira: If you need to ignore an exception, simply consume it blindly:
```py
try:
  raise Exception()
except:
  pass
```

judepereira:
@TonyRice Thanks, I'm updating a few things at my end, so avoid pushing to this branch until then.

asyncy/reporting/agents/SentryAgent.py
```diff
@@ -1,49 +1,49 @@
...
-                'app_name': reporting_event.app_name,
-                'app_version': reporting_event.app_version,
-                'story_name': reporting_event.story_name,
-                'story_line': reporting_event.story_line
-            })
-            # we utilize captureMessage because captureException
-            # will not properly work 100% of the time
-            # unless this is always called within try/catch block
-            self._sentry_client.captureMessage(message=err_str)
+            self._sentry_client.captureException(exc_info=exc_info)
```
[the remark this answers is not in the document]
judepereira: Cool, thanks!

judepereira:
Deploying on staging.

PR #371, opened 2019-07-05
Fix for ServiceOutputValidator

TonyRice:
This fixes the issue we discussed previously.

e.g.
```
runtime: An exception has occurred:
Failed to execute line: 'NoneType' object has no attribute 'get'
    at line 2: when client listen path: "/status" as request (in src/health.story)
    at line 4: redis get key: "__health_key_non_existent_ok" (in src/health.story)
```


TonyRice:
> tests ;-)

@wilzbach should be all good now!

PR #374, opened 2019-07-07
change to asynchronous postgres driver asyncpg

chrisstpierre:


asyncy/Apps.py
```diff
@@ -302,28 +295,13 @@ def get(cls, app_id: str):
...
-                            loop)
-            except (psycopg2.InterfaceError, psycopg2.OperationalError):
-                glogger.error('Connection to the DB has failed. Exiting.')
-                # Because _thread.interrupt_main() doesn't work.
-                os.kill(os.getpid(), signal.SIGINT)
-                break
+        conn = await Database.pg_conn(config)
+
+        await conn.add_listener('release',
+                                lambda _conn, _pid, _channel, payload:
```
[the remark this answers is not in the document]
chrisstpierre: So there was internal discussion here https://storyscript.atlassian.net/browse/PL-239 about why we should use something like `supervisor` or `circus`  instead, which is why I didn't include the kill logic here. But then again the pool manager should handle retry logic for connection problems, and I should be able to catch fatal exceptions in the callback somehow. I will add logic for that and update this.

asyncy/Config.py
```diff
@@ -13,9 +13,8 @@ class Config:
         'LOGGER_LEVEL': 'debug',
         'INGRESS_GLOBAL_STATIC_IP_NAME': 'storyscript-and-storyscriptapp',
         'APP_DOMAIN': 'storyscriptapp.com',
-        'POSTGRES': 'options='
-                    '--search_path=app_public,app_hidden,app_private,public '
-                    'dbname=postgres user=postgres',
+        'POSTGRES': 'postgres://postgres/asyncy?'
```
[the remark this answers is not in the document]
chrisstpierre: Its expected to be in the `libpq connection URI format`, but this is missing the options. I will make the update to `stack-kubernetes`

setup.py
```diff
@@ -38,7 +38,7 @@
         'storyscript==0.20.1',
         'ujson==1.35',
         'certifi>=2018.8.24',
-        'psycopg2==2.7.5',
+        'asyncpg==0.18.3',
```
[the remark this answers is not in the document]
chrisstpierre: I think that way too. But psycopg2 blocks the tornado event loop, even inside a coroutine. 1, 2

There are some other options on the Tornado Wiki, and some of them simply wrap psycopg2. But asyncpg is used in production, is supported, & is fast. So I think its a good choice.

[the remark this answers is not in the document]
judepereira: Just read about it a bit more, looks like we can use this for now. Since the changes are isolated, it'd be easy to revert too (just incase something doesn't work down the line).

asyncy/Apps.py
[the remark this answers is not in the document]
chrisstpierre: I added basic supervisor logic that polls every second and retries adding listener if there is a problem. Could you please review and tell me your thoughts on that?

asyncy/db/Database.py
```diff
@@ -1,103 +1,130 @@
...
-                             state: ReleaseState):
-        with cls.new_pg_cur(config) as db:
-            query = 'update releases ' \
-                    'set state = %s ' \
-                    'where app_uuid = %s and id = %s;'
-            db.cur.execute(query, (state.value, app_id, version))
-            db.conn.commit()
+    async def update_release_state(cls, glogger, config, app_id, version,
+                                   state: ReleaseState):
+        conn = await cls.pg_conn(config)
```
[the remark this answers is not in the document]
chrisstpierre: In my recent commit I made sure everything was in a `async with` context, which makes sure connections are released back to the pool

asyncy/db/Database.py
```diff
@@ -1,103 +1,130 @@
...
+            decoder=json.loads,
+            schema='pg_catalog',
+        )
+        await conn.set_type_codec(
+            'uuid',
+            encoder=str,
+            decoder=str,
+            schema='pg_catalog',
+        )
+        return conn
```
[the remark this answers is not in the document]
chrisstpierre: Also changed

chrisstpierre:
@judepereira This implementation only uses the config the first time it is provided. But all of the class methods accept a config, but the database connection config shouldn't change at any time right?

judepereira:
> @judepereira This implementation only uses the config the first time it is provided. But all of the class methods accept a config, but the database connection config shouldn't change at any time right?


No, the database config won't change at runtime.

asyncy/Apps.py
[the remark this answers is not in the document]
judepereira: Your basic state keeping looks okay. However, you're stealing one connection from the pool and holding on to it forever. Perhaps create your own connection and play with that? See https://github.com/MagicStack/asyncpg/issues/189#issuecomment-328569768 for manual connection open + close.

Connections should be taken from the pool only if they are going to be returned back.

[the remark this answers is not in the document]
judepereira: Also, please rename `_release_con` to something more appropriate.

asyncy/db/Database.py
```diff
@@ -1,103 +1,139 @@
...
-                             state: ReleaseState):
-        with cls.new_pg_cur(config) as db:
-            query = 'update releases ' \
-                    'set state = %s ' \
-                    'where app_uuid = %s and id = %s;'
-            db.cur.execute(query, (state.value, app_id, version))
-            db.conn.commit()
+        async with pool.acquire() as con:
+            # queries in transaction callback are rolled back on failure
+            async with con.transaction():
```
judepereira: No need for a transaction here. There's just one query.

asyncy/db/Database.py
```diff
@@ -1,103 +1,139 @@
...
-            json_object_keys(
-                (containerconfig->>'auths')::json
-            ) registry
-            from app_public.owner_containerconfigs
-            )
-            select name, containerconfig
-            from containerconfigs
-            where owner_uuid = %s and registry = %s
+    async def get_container_configs(cls, app, registry_url):
+        pool = await cls.pg_pool(app.config)
```
judepereira: Instead of repeating this line everywhere, can you create a wrapper which takes a connection from the pool directly and returns it? This will make it easier to test as well.

tests/unit/Apps.py
```diff
@@ -49,41 +43,64 @@ def foo(*args, **kwargs):
...
+@mark.asyncio
+async def test_supervise_listener(patch, db, magic,
+                                  config, logger, async_mock):
+    loop = magic()
+    patch.object(Apps, 'listen_to_releases', new=async_mock())
+    try:
+        await asyncio.wait_for(Apps.supervise_release_listener(config,
+                                                               logger, loop),
+                               timeout=3.2)
+    except asyncio.TimeoutError:
```
judepereira: Use `with pytest.raises(asyncio.TimeoutError)`. This will ensure that your exception was thrown.

tests/unit/Apps.py
```diff
@@ -49,41 +43,64 @@ def foo(*args, **kwargs):
...
+    patch.object(Apps, '_release_con')
+    patch.object(Apps._release_con, '_con', True)
+    patch.object(Apps._release_con, '_listeners', {'release'})
+
+    Apps.listen_to_releases.mock.call_count = 0
+    try:
+        await asyncio.wait_for(Apps.supervise_release_listener(config,
+                                                               logger, loop),
+                               timeout=1.2)
+    except asyncio.TimeoutError:
```
judepereira: Same.

asyncy/Apps.py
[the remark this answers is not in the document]
chrisstpierre: Sure I changed it to use a new con that is outside of the pool. Although I think its fine actually to use a connection from the pool even if its not returned, but if we are thinking of the pool as only for new queries than this makes sense.

asyncy/Apps.py
```diff
@@ -302,28 +301,30 @@ def get(cls, app_id: str):
...
-                            cls.reload_app(config, glogger, notify.payload),
-                            loop)
-            except (psycopg2.InterfaceError, psycopg2.OperationalError):
-                glogger.error('Connection to the DB has failed. Exiting.')
-                # Because _thread.interrupt_main() doesn't work.
-                os.kill(os.getpid(), signal.SIGINT)
-                break
+            await asyncio.sleep(1)
+            con = cls._release_con
+            if not con or not con._con or 'release' not in con._listeners:
```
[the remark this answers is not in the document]
chrisstpierre: I added release constants 🤷‍♂ anyways I know theres an entity but Im not sure if we want to expand on that later like an "orm" but I thought release constants is fine for now

asyncy/db/Database.py
chrisstpierre: yep. changed

asyncy/constants/ReleaseConstants.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+
+
+class ReleaseConstants:
+    table = 'release'
```
judepereira: This is not the table name, but rather the "notification event name". Perhaps rename to `notif_name`?

asyncy/Apps.py
```diff
@@ -309,28 +310,30 @@ def get(cls, app_id: str):
...
+                                   asyncio.run_coroutine_threadsafe(
+                                       cls.reload_app(config,
+                                                      glogger, payload),
+                                       loop))
+            cls.release_listener_db_con = con
+            glogger.info('Listening for new releases...')
+        except (OSError, asyncpg.exceptions.InterfaceError,
+                asyncpg.exceptions.InternalClientError):
+            # kill the server if the database listener is not listening
+            os.kill(os.getpid(), signal.SIGINT)
```
wilzbach: I understand that this is from the previous code, but isn't this a bit too drastic?

asyncy/Apps.py
```diff
@@ -309,28 +310,30 @@ def get(cls, app_id: str):
...
-                            loop)
-            except (psycopg2.InterfaceError, psycopg2.OperationalError):
-                glogger.error('Connection to the DB has failed. Exiting.')
-                # Because _thread.interrupt_main() doesn't work.
-                os.kill(os.getpid(), signal.SIGINT)
-                break
+            await asyncio.sleep(1)
+            con = cls.release_listener_db_con
+            if not con or not con._con or \
+                    ReleaseConstants.table not in con._listeners:
```
wilzbach: why do we need the second condition?

asyncy/db/Database.py
```diff
@@ -1,192 +1,211 @@
...
+
+            q1_vals = [(record['cpu_units'], record['memory_bytes'],
+                        record['service_uuid'], record['tag'])
+                       for record in data]
+
+            q2_vals = [(record['service_uuid'], record['tag'])
+                       for record in data]
+
+            await con.execute_many(query1, q1_vals)
+            await con.execute_many(query2, q2_vals)
```
wilzbach: while we are at it, we could rename the queries to sth. better than query1 and query2

tests/conftest.py
```diff
@@ -15,8 +15,20 @@ def magic(mocker):
...
+            return self.aenter
+
+        async def __aexit__(self, *args):
+            pass
+    return AsyncContextManagerMock(magic)
+
+
 @fixture
 def async_mock():
+
```
wilzbach: unrelated

tests/unit/Apps.py
```diff
@@ -292,13 +308,13 @@ def test_get(magic):
     )
 
     TooManyServices.__init__.assert_called_with(20, 15)
-    Database.update_release_state.assert_called()
+    Database.update_release_state.mock.assert_called()
```
wilzbach: Why do we need the `.mock`?

asyncy/Apps.py
judepereira: Naa, we don't want an engine which doesn't listen for new releases.

asyncy/db/Database.py
judepereira: Agreed.

tests/unit/Apps.py
judepereira: That's the way to access a mock object from an async mock.

asyncy/Apps.py
judepereira: not con: no connection
con._con: underlying connection (I am assuming this)
a not in con._listeners: notify listener lost in space

chrisstpierre: correct. a added comment to this effect

asyncy/db/Database.py
chrisstpierre: done

judepereira:
@chrisstpierre please refrain from force-pushing your commits. My commit which resolved the merge conflict is lost, and I had to do it all over again.

asyncy/Apps.py
wilzbach: Yes, but you're only interested in the listener being set (third condition). My point was that the third condition implies the second one.

judepereira: Oh, I don't think so. The listener could have been lost (not sure why though, but just keeping it for safety).

asyncy/Apps.py
wilzbach: But you kill the entire engine here, so no recovery from this is possible. This makes the supervision loop in `supervise_release_listener` a bit moot.

judepereira: I know, the monitoring logic is a little convoluted. The only effect it has is to ensure that there's a listener for that notification.

judepereira: Also, recovery is available at a lower level - k8s will restart the pod.

tests/unit/db/Database.py
```diff
@@ -127,91 +177,113 @@ def test_get_release_for_deployment(patch, config, database):
...
-    ]
-    database.conn.commit.assert_called()
-
-
-def test_get_service_by_alias(config, database):
+    await Database.update_service_usage(config, data)
+
+    pool.con.execute_many.mock \
+        .assert_any_call(query1,
+                         [(2, 100, '2614fee0-6b2a-4cd8-b4e6-6bbeab4eff84',
```
anukul: No reason to hardcode these values :confused:

asyncy/Apps.py
chrisstpierre: I think it is reasonable for now - at least until we integrate a process supervisor. Not intended to be a longterm solution

PR #375, opened 2019-07-09
Add float mutations

anukul:


anukul:
WIP: Tests for `is_nan` and `is_infinity`

anukul:
Should I add an `approx_equal` mutation? It's currently commented out in the language.

judepereira:
What does approx_equal mean? Two elements are either equal, or they aren't.

judepereira:
@anukul nice work :)

Quick one - most of the float ones, such as cos, sin, etc, are applicable to integers as well. What are your thoughts on allowing these for integers?

anukul:
> What does approx_equal mean? Two elements are either equal, or they aren't.

Not floats :P
https://www.python.org/dev/peps/pep-0485/#rationale
e.g.
```
$ python3
>>> from math import *
>>> tan(pi / 4) == 1
False
>>> isclose(tan(pi / 4), 1)
True
```
Recently had to use this in the service resource usage metrics [PR](https://github.com/storyscript/platform-engine/blob/service_usage/tests/unit/ServiceUsage.py#L183-L184) as well.

> most of the float ones, such as cos, sin, etc, are applicable to integers as well.

Yeah, I would expect `36.sqrt()`, `1.acos()` etc. to be possible.
@wilzbach @judepereira Should I add it in the language (and here)?

wilzbach:
> Should I add an approx_equal mutation? It's currently commented out in the language.

It would be `approxEqual` (Storyscript is camelCase) and yeah when dealing with floating-point we need this (though most users won't ever use it).

tl;dr: we should probably disallow `==` for floating point and enforce approxEqual.

> Quick one - most of the float ones, such as cos, sin, etc, are applicable to integers as well.
>> Yeah, I would expect 36.sqrt(), 1.acos() etc. to be possible.

Well integers implicitly cast to floats for most operations in the compiler, so this wouldn't be too unexpected.

anukul:
> Storyscript is camelCase

In that case, I see a bunch of snake_case [mutations](https://github.com/storyscript/storyscript/blob/master/storyscript/compiler/semantics/functions/HubMutations.py) that need fixing.

wilzbach:
> I see a bunch of snake_case mutations that need fixing.

:+1: (we only recently agreed to use camelCase everywhere for official Storyscript examples).

> Not sure when this is applicable

I said "operations" (e.g. `1 + 2.5` results in a `float`). This is currently implemented. We could introduce such a casting for mutations too, because this wouldn't be too unexpected or alternatively simply add the same list for ints.

judepereira:
@wilzbach can we do that for built-ins too? 

And yes, let's change stuff to camelCase. @anukul Can you create a subsequent PR for this?

wilzbach:
> @wilzbach can we do that for built-ins too?

We can do almost everything ;-)

The compiler currently uses a pre-built hashmap for to find the mutation overloading quickly (see e.g. https://github.com/storyscript/storyscript/blob/master/storyscript/compiler/semantics/functions/Mutation.py#L32

However we already do implicit casting for all other arguments, see e.g.

https://github.com/storyscript/storyscript/blob/master/storyscript/compiler/semantics/functions/Function.py#L54

(we just don't emit the `as` rewrites here yet, but In just opened an issue for it https://github.com/storyscript/storyscript/issues/993).

tl;dr: easier and faster to simply to add the mutations directly to int, but if there are other advantages to do implicit casting we could look in that too.

PR #378, opened 2019-07-10
Fix error: service alias can be None

anukul:


wilzbach:
TESTS!!!

judepereira:
@anukul let's add these in a follow up PR.

PR #381, opened 2019-07-11
Add support for deploying only staging/production/dev apps.

judepereira:
By default, all apps marked with the environment PRODUCTION in the db
will be deployed. All other apps, with other environments will not be
deployed. This helps with maintaining one engine per app_environment
in the same cluster.


judepereira:
Waiting on @chrisstpierre's PR for tonight to see if we can get it merged. If not, I'll merge this tomorrow.

PR #382, opened 2019-07-12
Upgrade to use Python 3.7.

judepereira:
Foundation for #374 (uses asynccontextmanager).


judepereira:
@wilzbach 
> Why not set python_requires='>=3.7' in the setup.py?

Done.

PR #385, opened 2019-07-12
Await on the database call.

judepereira:


judepereira:
Not interested in improving test coverage for ServiceUsage as @anukul will do so in his PR.

PR #388, opened 2019-07-12
Set release state to TEMP_DEPLOYMENT_FAILURE on internal errors

anukul:
See https://github.com/storyscript/platform-engine/issues/386


judepereira:
@anukul database schema changes? Need to merge that first.

anukul:
@judepereira https://github.com/storyscript/database/pull/81

wilzbach:
> @anukul database schema changes? Need to merge that first.

That's why we should move the database schema in here ;-)

judepereira:
@anukul can you also investigate why https://codecov.io/gh/storyscript/platform-engine/pull/388/diff#D1-501 is a partial?

judepereira:
> That's why we should move the database schema in here ;-)

@wilzbach I don't think so. The db schema has a ton of RLS, which the engine has nothing to do with. If it were plain schema, it would be best to move it here.

anukul:
@judepereira coverage fixed. (`elif` -> `if`)

judepereira:
Isn't elif correct?

anukul:
Not for 100% coverage :p
That branch isn't executed when the exception is not `StoryscriptError`, so `elif` only ever returns `True`, making it a partial.

judepereira:
@anukul okay - waiting for the CLI to be rolled out.

PR #391, opened 2019-07-15
Implement continue

anukul:
This change will be present in the next Storyscript compiler release.


asyncy/processing/Lexicon.py
```diff
@@ -130,6 +130,16 @@ def _does_line_have_parent_method(story, line, parent_method_wanted):
             # There is no parent, this is an illegal usage of break.
             raise InvalidKeywordUsage(story, line, 'break')
 
+    @staticmethod
+    async def continue_(logger, story, line):
+        # Ensure that we're in a foreach loop. If we are, return CONTINUE,
+        # otherwise raise an exception.
+        if Lexicon._does_line_have_parent_method(story, line, 'for'):
```
wilzbach: What about `while`?

tests/integration/processing/Lexicon.py
```diff
@@ -261,6 +261,17 @@ def __init__(self, cases, preparation_lines=''):
...
     ),
+    TestSuite(
+        preparation_lines='a = [1, 1, 1, 2, 3, 4, 5]\n'
+                          'b = 0\n',
+        cases=[
+            TestCase(append='foreach a as elem\n'
+                            '   if elem % 2 == 0\n'
+                            '       continue\n'
+                            '   b = b + elem\n',
+                     assertion=ContextAssertion(key='b', expected=11))
```
wilzbach: Needs a test case for `while` too.

anukul:
@seb There is no `while` yet in the engine.
I created a separate issue, and will probably get to it tomorrow.

judepereira:
@wilzbach can we lower foreach with while? then the engine no longer needs to support both while and foreach

wilzbach:
> @wilzbach can we lower foreach with while? then the engine no longer needs to support both while and foreach

Yes, I thought about this a while ago too. See e.g. https://github.com/storyscript/storyscript/issues/703

anukul:
@judepereira @wilzbach ready for review

tests/integration/processing/Lexicon.py
judepereira: @anukul test for while?

anukul: There is no `while` in the engine yet.

judepereira: Oh, okay then.

Issue #397, opened 2019-07-23
No way to create directory using the file service


TonyRice:
Currently, the file service does not allow you to create a directory.

PR #398, opened 2019-07-24
adding support for mkdir, and mkdirp in the file service

TonyRice:
See issue #397


asyncy/processing/internal/File.py
```diff
@@ -24,6 +25,36 @@ def safe_path(story, path):
     return f'{story.get_tmp_dir()}{os.fspath(path)}'
 
 
+def mkdir_error(e: IOError, path, story, line):
+    if e.errno == errno.EEXIST and os.path.isdir(path):
```
wilzbach: Use the `exist_ok` instead ;-)

asyncy/processing/internal/File.py
```diff
@@ -24,6 +25,36 @@ def safe_path(story, path):
...
+    path = safe_path(story, resolved_args['path'])
+    try:
+        os.mkdir(path)
+    except IOError as e:
+        mkdir_error(e, path, story, line)
+
+
+@Decorators.create_service(name='file', command='mkdirp', arguments={
+    'path': {'type': 'string'}
+})
```
wilzbach: We want to abstract complexity for the user. This shouldn't be a separate method. I think we should do this by default.

TonyRice: That's what I was initially thinking, I'll go ahead and do that

asyncy/processing/internal/File.py
TonyRice: Good thinking. Makes life easier.

asyncy/processing/internal/File.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import errno
```
wilzbach: Unused

TonyRice: Removed!

Issue #335, opened 2019-06-12
HTTP service can only handle utf8-text


wilzbach:
A user should be able to upload/download images or other files.

PR #399, opened 2019-07-24
Allow the file service to read/write raw bytes

TonyRice:
This is useful for when receiving a raw file via the http service. The raw argument can be something else. Completely open to any suggestion for that. This is related directly to #335 

It makes the following possible:
```
http server as client
    if r.file and r.file["body"]
        file write path:"/file.zip" content: r.file["body"]
    r write content: (file read path:"/file.zip" raw: true)
```
```
curl -H "Host: a-a-1.storyscriptapp.com" -X POST -F "file=@hello.zip" http://127.0.0.1:8888/ --output myfile.zip
```


TonyRice:
> LGTM, but could we add your full example to the testsuite to ensure the engine handles bytes arrays correctly?

Can you elaborate more on what you mean? I'm not entirely sure what you mean

wilzbach:
> I'm not entirely sure what you mean. I'm probably thinking too hard :D

I was talking about the example that you posted in the PR description. We should add it to the testsuite.

TonyRice:
> > I'm not entirely sure what you mean. I'm probably thinking too hard :D
> 
> I was talking about the example that you posted in the PR description. We should add it to the testsuite.

It should be in there unless I'm missing something entirely.

TonyRice:
@judepereira I have https://github.com/storyscript/file updated, I just need to get pr #401 and #402 wrapped up since the update in the file readme includes using the http service to receive raw bytes.

wilzbach:
> It should be in there unless I'm missing something entirely.

Uhm so I was referring to the Storyscript example being added as integration test, i.e. making sure the whole piece works as expected whereas you just add two unittests here which only test the individual subparts.

TonyRice:
> > It should be in there unless I'm missing something entirely.
> 
> Uhm so I was referring to the Storyscript example being added as integration test, i.e. making sure the whole piece works as expected whereas you added two unittests here which only test the individual subparts.

Okay, that's what I thought you meant but wasn't fully sure. I can definitely do that. I'd like to do that in a separate PR so I can get these changes into master

PR #401, opened 2019-07-24
Support for proper type conversion within context data

TonyRice:
Since FileFormField is a namedtuple, it's not possible to access data stored within the field using the keys. This makes it possible to access FileFormField information in the following manner:

```
http server as client
    if r.file and r.file["body"]
        file write path:"/file.zip" content: r.file["body"]
    r write content: r.file["body"] 
```

Related to #335


TonyRice:
> I don't think this is a good idea to put such service specialization in the Resolver. Can't we convert to a dict beforehand? Also, this lacks tests.

I'm not exactly sure where things break down the line, but my next solution would do a context cleanup on story.prepare() where the context is initially set. This would make more sense to me, and it helps us stay out of the resolver. If that makes sense to you, I'll push that up along with the tests.

tests/unit/Story.py
```diff
@@ -234,14 +239,60 @@ def test_story_prepare(story):
...
+    patch.object(story, 'cleanup_context')
     story.app = app
     context = {'app': app.app_context}
     story.prepare(context=context)
+    story.cleanup_context.assert_called_with(context)
     assert story.environment == app.environment
     assert story.context == context
 
 
+def test_story_cleanup_context(patch, story, app):
```
TonyRice: @wilzbach please let me know if the tests should be improved, or changed

TonyRice: Not entirely sure the best way to test a recursive function like this

wilzbach: You assign the function to a variable before patching it (copy). Then, you can call you unpatched copy, but all recursive calls will go to the patched one.

wilzbach:
@TonyRice my point was that we should do this `cleanup_context` higher up in the hierarchy while/before we assemble the initial context. The main motivations are: (1) performance (no need to run  `cleanup_context`  every time a function is called) and (2) separation of concern. The interpreter shouldn't never know what a `FileFormField` is. It should only concern itself with the types it supports (e.g. dicts). Otherwise, we end up with a massive `cleanup_context` every time we add a new feature.

TonyRice:
> @TonyRice my point was that we should do this `cleanup_context` higher up in the hierarchy while/before we assemble the initial context. The main motivations are: (1) performance (no need to run `cleanup_context` every time a function is called) and (2) separation of concern. The interpreter shouldn't never know what a `FileFormField` is. It should only concern itself with the types it supports (e.g. dicts). Otherwise, we end up with a massive `cleanup_context` every time we add a new feature.

As of right now, cleanup_context is only called when a new story is created. The performance impact is very minimal. To move up the hierarchy, the next step above that would to just remove something like FileFormField, And FormField all together otherwise. Of course, we can run it right before we set the context, but the performance impact would be exactly the same.

story.prepare is ran immediately after the story is created. Before, I was doing something like this inside the Resolver which definitely made less sense.

wilzbach:
> As of right now, cleanup_context is only called when a new story is created. The performance impact is very minimal. 

I disagree:

https://github.com/storyscript/platform-engine/blob/77fb1bb97521315aa475a6523014659381ba390d/asyncy/Story.py#L273-L276

https://github.com/storyscript/platform-engine/blob/8b90adf887ab88fafca62558c9fa635b69c105c3/asyncy/processing/Lexicon.py#L76-L101

> To move up the hierarchy, the next step above that would to just remove something like FileFormField, And FormField all together otherwise.

Yes, as mentioned I don't see a reason why such details should be leaked into the Story context.

TonyRice:
Thanks for pointing that out! I did not notice that. I'll make some changes

asyncy/Story.py
```diff
@@ -273,6 +277,24 @@ def set_context(self, context):
...
 
-    def prepare(self, context=None):
+    def prepare(self, context: dict = None):
+        self.cleanup_context(context)
         self.set_context(context)
         self.environment = self.app.environment or {}
+
+    @staticmethod
+    def cleanup_context(context: dict):
+        if context is not None:
```
wilzbach: Insert a return here, then you can save one level of indentation.

asyncy/http_handlers/StoryEventHandler.py
```diff
@@ -36,9 +35,12 @@ class StoryEventHandler(BaseHandler):
...
             tf = self.get_req().files[key][0]
-            f = FileFormField(name=key, body=tf.body,
-                              filename=tf.filename,
-                              content_type=tf.content_type)
+            f = {
+                'name': key,
+                'body': tf.body,
+                'filename': tf.filename,
+                'content_type': tf.content_type
+            }
```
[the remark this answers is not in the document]
wilzbach: Also, won't this break:

 https://github.com/storyscript/platform-engine/blob/2570b8d11799ce5b17730b26d3fa8d0e7ca7f835/asyncy/processing/Services.py#L446-L448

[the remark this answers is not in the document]
TonyRice: The problem is that it'll always look like a tuple to the engine and serialization, and overloading anything related to the NamedTuple, adds a whole layer of complexity that isn't necessary. Our next step, is to just revert it back to FileFormField, and then call cleanup_context immediately after that, making FileFormField only exist for cosmetic purposes, at least in the StoryEventHandler which is what I did have before I made the commit. I'll push up that change and let me know what you think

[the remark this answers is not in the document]
wilzbach: I think it's time that @judepereira chimes in too. Don't have a strong opinion here.

storyruntime/Story.py
```diff
@@ -276,3 +280,22 @@ def set_context(self, context):
...
         self.set_context(context)
         self.environment = self.app.environment or {}
+
+    @staticmethod
+    def cleanup_context(context: dict):
+        if context is None:
+            return
+
+        for key, value in context.items():
+            if isinstance(value, CaseInsensitiveDict):
```
judepereira: Why are you doing this for a case insensitive dictionary? In the context, `app.secrets` is a case insensitive dictionary, to allow for case insensitive access to secrets.

storyruntime/Story.py
```diff
@@ -276,3 +280,22 @@ def set_context(self, context):
     def prepare(self, context=None):
         self.set_context(context)
         self.environment = self.app.environment or {}
+
+    @staticmethod
+    def cleanup_context(context: dict):
```
judepereira: I don't think this cleanup_context is the solution to accessing named tuples. The cleaner approach is to add support for accessing keys for namedtuples in the Resolver, since this will:
1. Save on memory and object creation all the time by duplicating maps
2. Will not need any modifications to the context
3. Support for nested tuples is automatically added in the right manner - I do see that you're handling this explicitly, but I don't think this is the best approach - converting the tuple to a dictionary. The reason to preserve the types in the context is because one might inject things in the context on the fly, and expect it to be a namedtuple later down the execution path

TonyRice: I originally added it within the Resolver and it did work pretty well. I'll re-add that

judepereira: Thanks :)

asyncy/http_handlers/StoryEventHandler.py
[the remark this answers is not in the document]
judepereira: Moving the proposed change to the resolver will prevent this from breaking, and other related things too.

asyncy/Story.py
```diff
@@ -273,6 +277,26 @@ def set_context(self, context):
         # Optimise this later.
         self.context['app'] = self.app.app_context.copy()
 
-    def prepare(self, context=None):
+    def prepare(self, context: dict = None):
+        self.cleanup_context(context)
```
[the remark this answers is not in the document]
judepereira: Moving the proposed change to the Resolver.

storyruntime/Story.py
TonyRice: adding tests

storyruntime/utils/Resolver.py
```diff
@@ -47,7 +48,13 @@ def path(cls, paths, data):
                     item = cls.range(path['range'], item, data)
                 else:
                     resolved = Resolver.object(path, data)
-                    item = item[resolved]
+                    # allow a namedtuple to use keys or index
```
judepereira:
```suggestion
                    # Allow a namedtuple to use keys or index
```

PR #402, opened 2019-07-24
Http response serialization

TonyRice:
Currently the http service has multiple serialization issues when responding with certain data. More specifically, if one were to create an "echo" http service responding with the request data like the following:

```
http server as client
    r write content: (r) 
```

It would throw a utf-8 error given the fact that ujson cannot serialize the bytes that would be included in the FileFormField. 

With this fix, the above would return something like this:

```
curl -H "Host: a-a-1.storyscriptapp.com" -X POST -F "file=@hello.zip" http://127.0.0.1:8888/ 

{"uri": "/", "path": "/", "headers": {"Host": "a-a-1.storyscriptapp.com", "User-Agent": "curl/7.58.0", "Accept": "*/*", "Content-Length": "381", "Content-Type": "multipart/form-data; boundary=------------------------3d71a0cc543ef7d8"}, "query_params": {}, "file": {"name": "file", "body": "UEsDBAoAAAAAAOOw2k7j5ZWwDAAAAAwAAAAJABwAaGVsbG8udHh0VVQJAAPaMhRd3TIUXXV4CwABBOgDAAAE6AMAAEhlbGxvIFdvcmxkClBLAQIeAwoAAAAAAOOw2k7j5ZWwDAAAAAwAAAAJABgAAAAAAAEAAACkgQAAAABoZWxsby50eHRVVAUAA9oyFF11eAsAAQToAwAABOgDAABQSwUGAAAAAAEAAQBPAAAATwAAAAAA", "filename": "hello.zip", "content_type": "application/octet-stream"}}
````

This solves that problem and the serialization of CaseInsensitiveDicts too.

There are a few PRs currently open related to #335

### CODE REVIEW

File: asyncy/processing/Services.py [hunk 1]
```diff
@@ -24,6 +27,35 @@
 from ..utils import Dict
 from ..utils.HttpUtils import HttpUtils
 from ..utils.StringUtils import StringUtils
+from ..utils.TypeUtils import TypeUtils
+
+
+class HttpDataEncoder(json.JSONEncoder):
```
TonyRice: If there's a better home for this please definitely let me know. Will add tests in the morning!

wilzbach: Needs a doc string

asyncy/processing/Services.py
```diff
@@ -24,6 +27,35 @@
...
+    def default(self, obj):
```
wilzbach: docstring

asyncy/processing/Services.py
```diff
@@ -24,6 +27,35 @@
...
+        elif isinstance(obj, CaseInsensitiveDict):
+            # convert this to a regular dict
+            return dict(obj.items())
+
+        return json.JSONEncoder.default(self, obj)
+
+    def encode(self, o):
+        return json.JSONEncoder.encode(self, self.__convert_types(o))
+
+    def __convert_types(self, o):
```
wilzbach: double underscore is reserved for Python internal. Don't mess with it. A single `_` is enough.

asyncy/processing/Services.py
```diff
@@ -24,6 +27,35 @@
...
+        if isinstance(o, dict):
+            for key, value in o.items():
+                # always convert a named tuple to a
+                # dictionary value so we can decode
+                # it properly with the keys included.
+                if TypeUtils.isnamedtuple(value):
```
wilzbach: Why don't we check for the exact named tuples that we support instead?

TonyRice: I think it would make sense to utilize the Story.cleanup_context function in pr #401 effectively shrinking this

PR #403, opened 2019-07-25
Add global context to App

TonyRice:
Currently, you cannot access global variables due to the context not being carried down to any child execution. For example, within the HTTP Service, we trigger handlers with a fresh context.

```
context = {
    ContextConstants.service_event: event_body,
    ContextConstants.server_io_loop: io_loop,
    ContextConstants.server_request: self
}
```

I simply define app.global_context, and pass it along to Stories.run. This may not be ideal, but it solves the problem where something like the code below returns None:

```
response = "Hello World"
http server as client
  when client listen method:"get" path:"/" as r
    log info msg: "{response}"
    r write content: response
```

### DIFF
--- a/asyncy/App.py
+++ b/asyncy/App.py
@@ -79,6 +79,7 @@ class App:
             'hostname': f'{self.app_dns}.{self.config.APP_DOMAIN}',
             'version': self.version
         }
+        self.global_context = {}
 
     def image_pull_policy(self):
         if self.always_pull_images is True:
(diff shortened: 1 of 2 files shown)


judepereira:
@TonyRice Perhaps shallow copy the map? We don't want values changing accidentally (although the compiler shouldn't allow this).

TonyRice:
Passing this off to @anukul since he's working on conflicting changes

PR #411, opened 2019-07-28
Undo: remove b16-service-name label

anukul:
Fixes https://github.com/storyscript/platform-engine/issues/410


tests/unit/Kubernetes.py
```diff
@@ -609,6 +609,7 @@ def test_new_ssl_context():
     tag = image.split(':')[-1]
     service_uuid = '85b6735f-e648-439e-bb18-9b4ad130f69f'
     service_tag_uuid = 'e5103d54-8a49-4619-9519-04a892dd6817'
+    b16_service_name = base64.b16encode('alpine'.encode()).decode()
```
wilzbach: use the actual value here.

asyncy/Kubernetes.py
```diff
@@ -567,6 +567,8 @@ def get_liveness_probe(cls, app, service: str):
 
         liveness_probe = cls.get_liveness_probe(app, service_name)
 
+        b16_service_name = base64.b16encode(service_name.encode()).decode()
```
wilzbach: Out of interest: any reason why we can't use full Base64? 
(blows up the result x4)

judepereira: Because of k8s value limitations.

judepereira:
Closing in favour of solving this the right way: https://github.com/storyscript/logstreamer/pull/6

PR #412, opened 2019-07-29
Throw error invalid key

judepereira:


asyncy/utils/Resolver.py
```diff
@@ -52,7 +52,12 @@ def path(cls, paths, data):
         except IndexError:
             raise StoryscriptRuntimeError(
                 message=f'List index out of bounds: {resolved}')
-        except (KeyError, TypeError):
+        except KeyError:
+            raise StoryscriptRuntimeError(
+                message=f'Map does not contain the key "{resolved}". '
+                f'Use map.get(key: <key> default: <default value>) to '
+                f'prevent an exception from being thrown.')
```
wilzbach: Might want to mention `map.contains` too.

asyncy/utils/Resolver.py
```diff
@@ -52,7 +52,12 @@ def path(cls, paths, data):
         except IndexError:
             raise StoryscriptRuntimeError(
                 message=f'List index out of bounds: {resolved}')
-        except (KeyError, TypeError):
+        except KeyError:
+            raise StoryscriptRuntimeError(
+                message=f'Map does not contain the key "{resolved}". '
+                f'Use map.get(key: <key> default: <default value>) to '
```
wilzbach: Can't we use the actual path name here. Would be a lot more user-friendly as they then just copy/paste it.

judepereira: Nope - due to recursion + fake lines, it might not always be feasible to get the actual path.

asyncy/utils/Resolver.py
judepereira: https://github.com/storyscript/runtime/pull/426

asyncy/utils/Resolver.py
wilzbach: We might be able to do this in the future soon though, e.g. https://github.com/storyscript/storyscript/issues/1062 and https://github.com/storyscript/storyscript/issues/593.

PR #414, opened 2019-07-30
Adding pattern type conversion to HttpDataEncoder

TonyRice:


tests/unit/processing/Services.py
```diff
@@ -885,7 +886,8 @@ def test_http_data_encoder(patch):
         'casedict': CaseInsensitiveDict(data={
             'key': 'value'
         }),
-        'namedtuple': namedtuple_obj(key='value')
+        'namedtuple': namedtuple_obj(key='value'),
+        'regex': re.compile('/foo/i')
```
wilzbach:
```suggestion
        'regex': re.compile('/foo/i'),
```

Otherwise, you continue to mess up the git blame when you touch unrelated lines ;-)

PR #418, opened 2019-07-31
Rename Package namespace

TonyRice:
This is currently a work in progress. It looks like renaming the package to storyscript will cause a conflict with the storyscript language package. I propose we rename this package to storyengine, or something similar. Any thoughts and opinions would be appreciated.


wilzbach:
> . I propose we rename this package to storyengine, or something similar. Any thoughts and opinions would be appreciated.

An alternative would be `storyscript.engine`, but I like short names too.


Other ideas:
- `storyruntime`
- `storyre` (runtime environment)
- `sre`

Engine was always a weird and not very fitting name to me.

TonyRice:
I definitely like storyruntime

judepereira:
+1 on storyruntime. 


Cheers,
Jude Pereira

PR #419, opened 2019-07-31
Latest changes into production

TonyRice:


wilzbach:
@TonyRice you might want to consider using merge commits for such updates as it will avoid many potential merge conflicts.
See also: https://git-scm.com/docs/git-merge

I personally use this small script for my setups: https://github.com/wilzbach/git-tools/blob/master/merge_stable, though just a normal `git merge` is good enough.

TonyRice:
Thanks for the suggestion!

Issue #329, opened 2019-06-12
Stacked contexts


wilzbach:
A `when`, `foreach`, `while` or listening to a service should result in a nested `context`. Defined variables SHALL not be visible after leaving the scope.

wilzbach:
Repeating the points I made 

- you can access to the parent scope variables
- you can write to the parent scope variables (if they are not readonly)
- you can't define a variable from the parent scope (that would make things ambiguous, you'll get an error)
- the variables that you define in your scope are gone after you leave the scope

The idea here is to force the user to be very explicit about his code.

```coffee
c = 0 # we made this explicit, it won't work otherwise
foreach arr as e
  c += e
```

The same applies to try/catch. That being said there's one exception: if statements. If you define a variable in all statements, then it will be hoisted to the parent scope. However, if you define it in only the if-block, but not the else-block, you'll get a nice compiler error (though that last part isn't fully implemented yet - see storyscript/storyscript#1080 for details). We could apply a similar logic to try/catch.

How to do:
---

Every time you enter a block, you create a new context with ref to the parent, once you're out of the block, you pop the current context.
For variable lookup, you look in your current scope and if you can't find anything, you delegate to parent scope. There are smarter ways, but that one should be easy to implement.

About the if special case: Compiler could either mark these variables specially or could insert a (dummy) init assignment in the parent scope.

PR #422, opened 2019-08-01
Adding try catch support

TonyRice:
The goal is to add support for basic exception handling as outlined here. https://docs.storyscript.io/storyscript/advanced/#exception-handling


storyruntime/processing/Lexicon.py
```diff
@@ -236,6 +236,34 @@ def unless_condition(logger, story, line):
...
 
+    @staticmethod
+    async def try_catch(logger, story, line):
+        from . import Stories
+        next_line = story.next_block(line)
+
+        if next_line is None:
+            raise StoryscriptError(
+                message=f'Try requires catch block',
+                story=story, line=line)
```
wilzbach: The compiler could/should enforce this if we want to make this illegal (there's a point for allowing it which is why it's currently valid).

storyruntime/processing/Lexicon.py
```diff
@@ -236,6 +236,34 @@ def unless_condition(logger, story, line):
...
+
+        try:
+            await Stories.execute_block(logger, story, line)
+        except BaseException as e:
+            output = next_line['output'][0]
+            try:
+                story.context[output] = str(e)
```
wilzbach: Uhm at the team summit we decided to not expose the exception message (or other context) yet to the user, but only allow him to differentiate between exceptions (user errors, recoverable, e.g. 404), and errors (application error, not recoverable, e.g. 500).

More details: https://github.com/storyscript/storyscript/issues/954

storyruntime/processing/Lexicon.py
```diff
@@ -236,6 +236,34 @@ def unless_condition(logger, story, line):
             return line['exit']
         return line['enter']
 
+    @staticmethod
+    async def try_catch(logger, story, line):
+        from . import Stories
```
wilzbach: I recommend refactoring the code so that you don't need this import here. One reason is that it will make unittest testing this harder (can't easily patch the class).

TonyRice: Funny you mention that I actually do have a refactor currently in progress. I was thinking it makes more sense to put execute_block and execute_line within the Lexicon. This completely avoids the inline imports altogether.

storyruntime/processing/Lexicon.py
TonyRice: I'd love to sync up on this and get a full breakdown. I'll review that issue, and be sure to ping you with any questions.

TonyRice: We're going to focus on implementing basic try catch as the docs currently outline. We can re-visit this in the future.

storyruntime/processing/Lexicon.py
```diff
@@ -325,6 +327,59 @@ def unless_condition(logger, story, line):
             return line['exit']
         return line['enter']
 
+    @staticmethod
+    async def try_catch(logger, story, line):
+        next_line = story.next_block(line)
```
judepereira:
```suggestion
        """Executes the try/catch/finally construct. If any StoryscriptError exception is thrown by the try block, the catch block will be invoked. However, if the error is not of type StoryscriptError, then it will be thrown up directly - in this case, the finally block will not be executed either (since the error that occurred is not a StoryscriptError, but rather a programming error in the runtime).
        
        :return: If a finally block is present, returns the line number of the finally block, otherwise, the line to be executed immediately after the catch block."""
        next_line = story.next_block(line)
```

storyruntime/processing/Lexicon.py
```diff
@@ -325,6 +327,59 @@ def unless_condition(logger, story, line):
...
+
+        if next_line is None:
+            return None
+
+        try:
+            await Lexicon.execute_block(logger, story, line)
+        except BaseException as e:
```
judepereira: Only StoryscriptError exceptions should be subjected to Storyscript try/catch blocks. All other errors are a result of incorrect runtime behaviour.
```suggestion
        except StoryscriptError as e:
```

storyruntime/processing/Lexicon.py
```diff
@@ -325,6 +327,59 @@ def unless_condition(logger, story, line):
...
+            if next_line['method'] == 'finally':
+                return next_line['enter']
+
+            line_output = next_line.get('output', [])
+
+            if line_output is not None and \
+                    len(line_output) > 0:
+                output = line_output[0]
+                if isinstance(e, StoryscriptError):
```
judepereira: Since we'll be catching StoryscriptError exceptions only, this entire if/else block can be replaced with `story.context[output] = str(e)`.

storyruntime/processing/Lexicon.py
```diff
@@ -325,6 +327,59 @@ def unless_condition(logger, story, line):
...
+                            line=line,
+                            root=e
+                        )
+                    )
+            else:
+                output = None
+
+            try:
+                await Lexicon.execute_block(logger, story, next_line)
+            except BaseException as exc:
```
judepereira: No need for this `except` block, since all you're doing is throwing it back.

storyruntime/processing/Lexicon.py
```diff
@@ -325,6 +327,59 @@ def unless_condition(logger, story, line):
...
+                # let's throw the error that happened
+                # inside the catch block
+                raise exc
+            finally:
+                if output is not None:
+                    del story.context[output]
```
judepereira: I see what you're doing here. Once the execution is complete, you're removing the exception object from the context. This works, and is a good approach for now.

However, an even finer approach (which will be future proof is to implement something like a COW context, where a an "overlay" context frame is created, which will automatically reclaim all variables created in the current context - foreach, while, try/catch, if/else, etc). Don't confuse this with stack frames though, since entering a foreach loop doesn't add to the stack. Perhaps create a new issue for this, and let it percolate in both our minds?

storyruntime/processing/Lexicon.py
```diff
@@ -329,59 +329,91 @@ def unless_condition(logger, story, line):
...
+                last_block = story.next_block(last_block)
+
+            return Lexicon.line_number_or_none(last_block)
+
+        def clean_output(_output, original_value):
+            if _output is not None and \
+                    _output in story.context:
+                del story.context[_output]
+                if original_value is not None:
+                    story.context[_output] = original_value
```
wilzbach: This looks suspicious like the wrong thing to do. You need to implement the concept of scopes and stack frames anyhow for other stuff.
stack frame:
- no access to parent scope (e.g. when a function gets called)
- save old context in a stack of stack frame and pop the current context when returning back from the function call

scope:
- have access to the parent scope
- BUT: variables written in one scope only last in that scope

wilzbach: There already is an issue for this: https://github.com/storyscript/runtime/issues/329

storyruntime/processing/Lexicon.py
```diff
@@ -329,59 +329,91 @@ def unless_condition(logger, story, line):
...
+        the catch block or finally block.
+        """
         next_line = story.next_block(line)
 
         if next_line is None:
             return None
 
+        async def next_block():
+            """
+            This will execute if the next block is a finally block.
```
wilzbach: It's a very bad practice to give a pretty specialized function such a generic name as it misleads people

storyruntime/processing/Lexicon.py
wilzbach: So that means we don't support access to the error object, right and this should be removed.
I'm aware that https://docs.storyscript.io/storyscript/advanced/#exception-handling mentions `as error`, but we _never_ defined how that object should look like.

storyruntime/processing/Lexicon.py
TonyRice: Is there a reason why we would not want to define variables to not be accessible after execution? Generally, a catch block would be able to define, and access globals variables. I'm just unsure, maybe I should move the discussion into #329 

In this specific case, I actually added the code due to a recent error I discovered in the compiler. The catch as error was only usable if I defined the error variable globally. This just sets it back.

storyruntime/processing/Lexicon.py
TonyRice: I'll rename it to something more specific

storyruntime/processing/Lexicon.py
TonyRice: If you take a look at this issue I just submitted, https://github.com/storyscript/storyscript/issues/1094

`as error` was working perfectly for me not too long ago, with the try catch implementation on the runtime of course. Trying to debug it, I realized I could simply define the variable globally to get the runtime to compile. When passing this to the context, I realized I was deleting it after the catch block executes. In the end, removing the global variable I defined. It's a bit of a weird situation, but the above is more or less a temporary fix to that problem.

I'm not sure if the compiler should throw an error since the variable was already defined, but I am experiencing an issue a bit intertwined in this.

storyruntime/processing/Lexicon.py
TonyRice: If you take a look at this issue I just submitted, https://github.com/storyscript/storyscript/issues/1094

`as error` was working perfectly for me not too long ago, with the try catch implementation on the runtime of course. Trying to debug it, I realized I could simply define the variable globally to get the runtime to compile. When passing this to the context, I realized I was deleting it after the catch block executes. In the end, removing the global variable I defined. It's a bit of a weird situation, but the above is more or less a temporary fix to that problem.

I'm not sure if the compiler should throw an error since the variable was already defined, but I am experiencing an issue a bit intertwined in this.

wilzbach: > Is there a reason why we would not want to define variables to not be accessible after execution? 

I was talking about the `as error`.
Though there are many other example, like `foreach` or `while`

> Generally, a catch block would be able to define, and access globals variables.

Definitely a big no on defining _global_ variables. You shouldn't be able to ever add new global variables. If you're talking about the parent scope, that's a no too.
Every block (indent) creates a new context with the following rules:
- you can access to the parent scope variables
- you can write to the parent scope variables (if they are not readonly)
- you can't define a variable from the parent scope (that would make things ambiguous, you'll get an error)
- the variables that you define in your scope are gone after you leave the scope

The idea here is to force the user to be very explicit about his code.

```coffee
c = 0 # we made this explicit, it won't work otherwise
foreach arr as e
  c += e
```

The same applies to try/catch. That being said there's on exception: `if` statements. If you define a variable in all statements, then it will be hoisted to the parent scope. However, if you define it in only the if-block, but not the else-block, you'll get a nice compiler error (though that last part isn't fully implemented yet - see https://github.com/storyscript/storyscript/issues/1080 for details). We could apply a similar logic to `try/catch`.

> In this specific case, I actually added the code due to a recent error I discovered in the compiler. 

Please don't try to workaround things unnecessarily. Ask e.g. on #storyscript as we very likely can fix things for you.
This just creates more tech debt. In this case though (I assume you're talking about https://github.com/storyscript/storyscript/issues/1094) it's only partially a compiler error as the code is incorrect according to your agreement on the team summit (=https://github.com/storyscript/storyscript/issues/954). Though of course it should give a better error message and say that `catch as error` is not allowed. Anyhow, let's continue that discussion on the respective issue.

wilzbach: > We could apply a similar logic to try/catch.

-> https://github.com/storyscript/storyscript/issues/1095

TonyRice: @wilzbach, unfortunately, I wasn't able to be a part of the majority of the discussions at the team summit since I was not there. That clarifies the majority of what I'm trying to figure out. It looks like this would now need an update to the compiler in order to continue implementing the try-catch. As it is right now, the compiler enforces the syntax ```catch as error```, but it throws an error when code within the catch block tries accessing it. I'm going to revert it to exactly what I had before.

storyruntime/processing/Lexicon.py
wilzbach: > This works, and is a good approach for now.

No it's a terrible approach and we shouldn't commit to such hacks.
Just consider:

```coffee
error = 1
try:
  crashFun()
catch as error:
  # foo

error += 1 # KABOOM
```

Yes, in same cases the compiler will forbid it statically, but what do we do here:

```coffee
try:
  crashFun()
catch as error:
   try:
     crashFun()
   catch as error:
     # ..
```

> (which will be future proof is to implement something like a COW context, where a an "overlay" context frame is created, which will automatically reclaim all variables created in the current context - foreach, while, try/catch, if/else, etc).

You need to implement this anyhow. Language depends on this https://github.com/storyscript/runtime/issues/329.

>  Perhaps create a new issue for this, and let it percolate in both our minds?

https://github.com/storyscript/runtime/issues/329 already exists for a long time and it's pretty simple: every time you enter a block, you create a new context with ref to the parent, once you're out of the block, you pop the current context. For variable lookup, you look in your current scope and if you can't find anything, you delegate to parent scope. There are smarter ways, but that one is possible in 10-20 lines.

storyruntime/processing/Lexicon.py
TonyRice: Trying the following code:

```
try
     throw "error"
catch
    log info msg: "Error Happened"
finally
     log info msg: "Finally"
```

Will result in
```
Error: syntax error in story at line 3, column 6

3|    catch
           ^

E0053: Unexpected end of line. Expected: ['_AS'].
```

wilzbach: > I wasn't able to be a part of the majority of the discussions at the team summit since I was not there. 

Yeah we should have kept better notes, but I tried to move all important points to GH. In this case it's https://github.com/storyscript/storyscript/issues/954.

> hat clarifies the majority of what I'm trying to figure out. It looks like this would now need an update to the compiler in order to continue implementing the try-catch. As it is right now, the compiler enforces the syntax catch as error

Sorry, yes, we apparently need to talk more.

> As it is right now, the compiler enforces the syntax catch as error

Ok -> https://github.com/storyscript/storyscript/issues/1096

storyruntime/processing/Lexicon.py
judepereira: @wilzbach yes - same approach. Let's defer that to a follow up PR perhaps? I wouldn't want this PR to be extended simply at all. Also, a follow up PR makes sense, since functions, foreach, and all other items would be touched upon as well. @TonyRice let's chat about this live, since I have some ideas on implementing this concept.

judepereira:
> @wilzbach Nested try/catch is possible.

By the looks of Tony's approach, nested try/catch will work just fine. @TonyRice please add appropriate integration tests in Lexicon.py.

tests/integration/processing/Lexicon.py
```diff
@@ -634,7 +634,75 @@ def __init__(self, cases, preparation_lines=''):
...
+                     assertion=RuntimeExceptionAssertion(
+                         exception_type=StoryscriptError,
+                         context_assertion=ContextAssertion(
+                             key='_finally', expected=True
+                         )
+                     )),
+        ]
+    ),
+    TestSuite(
+        preparation_lines='try\n'
```
judepereira: Need a test like this as well:
```
try
  try
    throw "x"
  catch
    log info msg: "caught"
catch
  log error msg: "this should never happen"
finally
  log info msg: "outer finally"
```

Replace the log info lines with variable assignments to an outer context, and verify state via the ContextAssertion.

Issue #327, opened 2019-06-12
Use custom data types for the interpreter


wilzbach:
Instead of using the plain python data types everywhere in the interpreter we want to use our own Storyscript engine types as this will allow a few crucial points:
1) Ensure all runtime operations will happen as defined (or error as defined). We can still use the Python types under the hood, but by using our own types we make sure that all operations are fully covered by the testsuite
2) Maintenance. We can ensure that e.g. all string operation code lies exactly in `String.py` and isn't distributed over a ton of methods.
3) Serialization and logging. This will make it a lot easier to differentiate between user-code variables and engine variables
4) Future-proof: we already have our own data types (e.g. `time`) which is currently unimplemented. With our custom variable type, implementing time would be a lot faster and easier as one would only need to add the type and then implement all operations on it
...

PR #437, opened 2019-08-06
TypeUtils.safe_type

TonyRice:
This is a basic cleanup/sanitization that is performed on the Resolver, and TypeResolver.check_type_cast

If the item in question is something that can potentially leak internal data, it will convert it to a type that doesn't leak that data. For example, a StreamingService will include a container id, and hostname. This data isn't something we would want to expose to someone's story. This solves that.


TonyRice:
Adding tests

storyruntime/utils/TypeUtils.py
```diff
@@ -10,3 +15,35 @@ def isnamedtuple(o):
...
+                service.commands[key] = TypeUtils.safe_type(
+                    value
+                )
+            return service
+        elif isinstance(o, InternalCommand):
+            return SafeInternalCommand(
+                arguments=o.arguments,
+                output_type=o.output_type
+            )
+        else:
```
wilzbach: Maybe a failsafe assert that we only have basic types by here to protect against future programming errors.

storyruntime/App.py
```diff
@@ -79,6 +79,7 @@ def __init__(self, app_data: AppData):
             'hostname': f'{self.app_dns}.{self.config.APP_DOMAIN}',
             'version': self.version
         }
+        self.global_context = {}
```
wilzbach: This change seems to be unrelated.

TonyRice: @wilzbach  I went ahead and merged a local branch I had that contained this commit. 399dffada8cf31a98d63411ab809029a9c5fbb24 It allows me to access global variables. @anukul is working on something related to that, but I think this could be put into its own PR, and merged in for a quick fix for accessing global variables

anukul: I don't think https://github.com/storyscript/runtime/commit/399dffada8cf31a98d63411ab809029a9c5fbb24 is sufficient though, see https://github.com/storyscript/runtime/commit/e386596333475d350fbbed10a07762e817c7cb6a.  
This will be changed soon though, with the PR for https://github.com/storyscript/platform-engine/issues/329 coming in shortly. I was waiting for some issues in the compiler to be fixed the last time I was working on this.

TonyRice: I'm going to be removing my change out of this pr. Can we cherry pick your global context into its own PR?

judepereira:
@TonyRice @wilzbach I'm not a fan of the approach taken in this PR, but perhaps I could be wrong. Since we'll constantly be adding new namedtuples, and other objects, it's very easy for the type safe mapper to be outdated. Instead of this, for any special type, say, namedtuples and objects, why not just print their name and object address instead of the contents? And then, if we need to, (like the FileFormField) exclusively mark these classes/namedtuples as serialisable towards the end user?

judepereira:
@TonyRice @wilzbach My approach can easily be implemented by not doing an implicit _as_dict here: https://github.com/storyscript/runtime/blob/master/storyruntime/processing/Services.py#L86

TonyRice:
> @TonyRice @wilzbach I'm not a fan of the approach taken in this PR, but perhaps I could be wrong. Since we'll be constantly adding new namedtuples, and other objects, it's very easy for the type safe mapper to be outdated.
> 
> Instead of this, for any special type, say, namedtuples and objects, why not just print their name and object address instead of their contents? And then, if we need to, (like the FileFormField) exclusively mark these classes/namedtuples as serialisable towards the end user?

That's definitely a simpler approach.

wilzbach:
> Since we'll be constantly adding new namedtuples, and other objects,

But these shouldn't be added into the lexicon. That was my point in a different issue that everything that isn't a primitive type will introduce more and more unwanted complexity into that part of the engine - especially because it should concern itself just with user types.

> it's very easy for the type safe mapper to be outdated.

Instead of this, for any special type, say, namedtuples and objects, why not just print their name and object address instead of their contents?

You should __never ever leak addresses__!
However, yes I already pointed out that we need a case to catch everything that isn't a primitive type or dict/array, but this should be be an `assert` that fails the program instead as you want to have clear control over what and how types get exposed to the user 

>  And then, if we need to, (like the FileFormField) exclusively mark these classes/namedtuples as serialisable towards the end user?

Yes, a special `storySerialize` method is a good idea. Though you should still crash the program if the user managed to print anything that we didn't foresee as that's a critical error and there's no good message to print to. Also, there aren't that many cases, it's basically just the streaming service object for now (and service outputs if you don't convert them to a dict before entering the lexicon).

storyruntime/utils/TypeUtils.py
TonyRice: What basic data types do you propose? I know we would want to at least assert for str, int, float, complex, bool, and dict

wilzbach: Well there are many reasons why https://github.com/storyscript/runtime/issues/327 is very important, but that's one of them ;-)

> complex, bool, and dict

- There's no complex in Storyscript
- There's no bool in Python
- BUT there's `list`

storyruntime/utils/Resolver.py
```diff
@@ -42,6 +42,10 @@ def path(cls, paths, data):
                 if isinstance(path, str):
                     item = item[path]
 
+                # Sanitize this item so we can ensure that
+                # any unwanted data doesn't leak.
+                item = TypeUtils.safe_type(item)
```
judepereira: @TonyRice This isn't the right method for this. Add the call to TypeUtils.safe_type to the method `resolve` (towards the bottom of the same file), since that is the gatekeeper for accessing data.

storyruntime/utils/TypeUtils.py
```diff
@@ -10,3 +19,44 @@ def isnamedtuple(o):
...
+            # ensure the type is a primitive type
+            assert isinstance(o, FileFormField) or \
```
judepereira: This will throw an assertion error, which might not be readable to the end user when it gets displayed in the logs. Instead of the assert, replace with an if/else or something like `allowed_types.contains(type(o))`? Raise a StoryscriptRuntimeError :)

storyruntime/utils/TypeUtils.py
```diff
@@ -1,5 +1,29 @@
...
+        FormField,
+        re.Pattern,
+        str,
+        int,
+        float,
+        bool,
+        list,
+        dict,
+        bytearray,
+        bytes
```
wilzbach: Can we add lexicon/integration tests for all of these types to ensure they properly convert to/from string?

storyruntime/processing/Services.py
```diff
@@ -88,7 +88,7 @@ def _convert_types(self, o):
                         'command': v.command
                     }
                 elif TypeUtils.isnamedtuple(v):
-                    o[k] = v._asdict()
+                    o[k] = type(v).__name__
```
wilzbach: Why would we want to leak the implementation name?

TonyRice: Maybe we should follow a similar approach to safe_type, and throw an error if it's a type we did not expect

storyruntime/utils/TypeUtils.py
TonyRice: Yes!

tests/integration/processing/Lexicon.py
```diff
@@ -229,6 +229,48 @@ def __init__(self, cases, preparation_lines=''):
                          key='a', expected='helloworld'))
         ]
     ),
+    TestSuite(
+        preparation_lines='list = ["hello", "world"]\n'
+                          'dict = {"hello": "world"}\n'
+                          'file write path: "/tmp.txt" content: "hello"\n'
+                          'bytes = file read path: "/tmp.txt" raw: true',
+        cases=[
+            TestCase(prepend='a = "foo is {true}"',
```
wilzbach: Why not just 
```suggestion
            TestCase(prepend='a = "{true}"',
```

or `true as string`

storyruntime/processing/Services.py
wilzbach: so we still should do this, no ;-)
e.g. with a simple `assert(0, ...)`.

PR #443, opened 2019-08-09
While loop implementation

TonyRice:
Basic while loop implementation, unit tests in progress.


storyruntime/processing/Lexicon.py
```diff
@@ -415,6 +416,34 @@ def unless_condition(logger, story, line):
...
+                break
+
+            result = await Lexicon.execute_block(logger, story, line)
+
+            call_count += 1
+
+            if result == LineSentinels.CONTINUE:
+                continue
+            elif result == LineSentinels.BREAK or \
+                    result == LineSentinels.RETURN:
```
judepereira: `return` must be treated differently. A `return` sentinel must not just break the while loop, but also return the `return` sentinel outside.

storyruntime/processing/Lexicon.py
```diff
@@ -450,6 +479,11 @@ def unless_condition(logger, story, line):
             assert len(args) == 0, \
                 'return may not be used with a value in a when block'
 
+            return LineSentinels.RETURN
+        elif cls._does_line_have_parent_method(story, line, 'while'):
+            assert len(args) == 0, \
+                'return may not be used with a value in a while block'
```
judepereira: Why? What if this `while` block is in a function, and that function returns something? There's no need for this check here.

tests/integration/processing/Lexicon.py
```diff
@@ -831,6 +831,54 @@ def __init__(self, cases, preparation_lines=''):
...
+        cases=[
+            TestCase(
+                append='while i < 10\n'
+                       '   i = i + 1\n'
+                       'outside = true',
+                assertion=[ContextAssertion(key='outside', expected=True),
+                           ContextAssertion(key='i', expected=10)])
+        ]
+    ),
+    TestSuite(
```
judepereira: Duplicate test of the above?

tests/integration/processing/Lexicon.py
```diff
@@ -831,6 +831,54 @@ def __init__(self, cases, preparation_lines=''):
...
+        preparation_lines='i = 0',
+        cases=[
+            TestCase(
+                append='while true\n'
+                       '   i = i + 1\n'
+                       'outside = true',
+                assertion=[ContextAssertion(key='outside', expected=True),
+                           ContextAssertion(key='i', expected=1000000)])
```
judepereira: Ummm, what's with the magic number 1000000? If you're intent is to prevent people from looping more than 1 million times, please throw a `StoryscriptRuntimeError`. At least that way their code breaks and doesn't execute randomly. Also, make this a constant somewhere and import it here.

tests/integration/processing/Lexicon.py
```diff
@@ -831,6 +831,54 @@ def __init__(self, cases, preparation_lines=''):
...
+                       '    a = 0\n'
+                       '    while a < 10\n'
+                       '      i = i + 1\n'
+                       '      a = a + 1\n'
+                       'outside = true',
+                assertion=[ContextAssertion(key='outside', expected=True),
+                           ContextAssertion(key='i', expected=100),
+                           ContextAssertion(key='a', expected=10)])
+        ]
+    ),
```
judepereira: Please add tests which test for `break` in a while loop, and `while` with `return` in a function.

storyruntime/processing/Lexicon.py
```diff
@@ -415,6 +416,34 @@ def unless_condition(logger, story, line):
...
+                not should_break:
+
+            # note this is only a temporary solution,
+            # and we will address this in the future.
+            if call_count >= 1000000:
+                logger.warn(
+                    'Call count limit reached within while loop. '
+                    'Only 1 million iterations allowed.'
+                )
+                break
```
wilzbach: Yeah as Jude mentioned this needs to be a non-recoverable error.

storyruntime/processing/Lexicon.py
wilzbach: Which should be verified with a test ;-)

tests/unit/processing/Lexicon.py
```diff
@@ -382,6 +382,111 @@ def test_lexicon_unless_false(logger, story, line):
...
+        story.next_block(story.tree['1'])
+    )
+
+    assert Lexicon.execute_line.call_count == 10
+    assert story.line_has_parent.call_count >= 10
+    assert story.resolve.call_count >= 10
+
+    story.resolve.assert_called()
+    story.line.assert_called()
+    story.line_has_parent.assert_called()
```
wilzbach: Not sure whether these white box tests is worth the hassle. JSON returned by compiler will evolve over time and the runtime too, so you'll continuously have to update this here.
Also, the compiler would give you most of the stuff you're dynamically checking for here. Please just open issues for all the things you think the runtime shouldn't need to figure out and we'll add it to the JSON tree.

TonyRice:
Currently investigating a performance-related issue that seems to lock up the runtime.

storyruntime/Apps.py
```diff
@@ -197,6 +197,8 @@ def make_logger_for_app(cls, config, app_id, version):
 
     @classmethod
     def get(cls, app_id: str):
+
+        print('APPSSS', cls.apps)
```
wilzbach: remove debug logging.

storyruntime/processing/Lexicon.py
```diff
@@ -427,6 +430,35 @@ def throw(logger, story, line):
         # Use story.next_block(line), because line["exit"] is unreliable...
         return Lexicon.line_number_or_none(story.next_block(line))
 
+    @staticmethod
+    async def while_(logger, story, line):
+        call_count = 0
+        while Resolver.resolve(line['args'][0], story.context):
+            # note this is only a temporary solution,
+            # and we will address this in the future.
```
wilzbach: maybe reference an issue about this (-> open issue s.t. it's not forgotten).

storyruntime/processing/Lexicon.py
```diff
@@ -427,6 +430,35 @@ def throw(logger, story, line):
...
+                raise StoryscriptRuntimeError(
+                    message='Call count limit reached within while loop. '
+                            'Only 100000 iterations allowed.',
+                    story=story, line=line
+                )
+
+            result = await Lexicon.execute_block(logger, story, line)
+
+            # Let's sleep so we don't take up 100% of the CPU
+            await asyncio.sleep(0.0002)
```
wilzbach: maybe only do this every tenth or one hundredth time to avoid too many (expensive) context switches.

tests/integration/processing/Lexicon.py
```diff
@@ -872,14 +872,61 @@ def __init__(self, cases, preparation_lines=''):
...
+                append='while true\n'
+                       '   i = i + 1\n',
+                assertion=RuntimeExceptionAssertion(
+                    exception_type=StoryscriptRuntimeError,
+                    context_assertion=ContextAssertion(
+                        key='i',
+                        expected=100000
+                    )
+                )
+            )
```
wilzbach: Can we add a test for sth. like:

```coffee
function foo returns boolean
   while true
       return true
```

storyruntime/processing/Lexicon.py
wilzbach: You could have referenced the issue here, s.t. it's easier to find it for future readers of this code.

TonyRice: Got it! See https://github.com/storyscript/runtime/issues/489

PR #448, opened 2019-08-12
Don't modify story during assignment

anukul:


wilzbach:
Tests?

anukul:
@seb updated, added test fails without this change.

judepereira:
@anukul can you explain?

anukul:
@judepereira [This](https://github.com/storyscript/runtime/blob/master/storyruntime/utils/Dict.py#L13) line modifies the story.

judepereira:
@anukul Can't we just fix that instead of deep copying all the time?

anukul:
@judepereira updated, thanks!

PR #456, opened 2019-08-13
Rename asyncy to storyscript

judepereira:


judepereira:
Ignore the Codecov coverage.

PR #457, opened 2019-08-13
Adding fix for decoding issue within http client

TonyRice:
Currently, doing something like retrieving google.com, will throw a decoding error. This effectively attempts to pull the charset from the headers and decode it properly


wilzbach:
Umm having actual http requests to the outside world in the testsuite isn't great for a few reasons:
- local development with offline or slow internet
- random changes/outages from these sites now affect the testsuite
- testsuite can't be run in total isolation mode anymore

What's your opinion about mocking out `http fetch` with fixed replies?

judepereira:
@wilzbach I hope the tests are in the integration directory. If they are, then it's okay. While we do have mocks in place, I like the idea of having real requests being made as part of the integration test suite.

judepereira:
For local development, running unit tests should always be < 10 seconds.

TonyRice:
Those are integration tests, I was going to use a story for those tests but just did that for a quick solution

PR #481, opened 2019-08-22
Cleanup tmpdir after App shutdown

TonyRice:
I will add integration tests for the entire file service in a separate PR.


storyruntime/App.py
```diff
@@ -255,3 +289,4 @@ def remove_subscription(self, sub_id: str):
         """
         await self.clear_subscriptions_synapse()
         await self.unsubscribe_all()
+        await self.cleanup_tmp_dir()
```
wilzbach: Might want to put all the other stuff in a try/catch to ensure that this is always done.

TonyRice: I shall do that

storyruntime/App.py
```diff
@@ -96,6 +99,37 @@ def image_pull_policy(self):
...
+
+        for root, dirs, files in os.walk(tmpdir, topdown=False):
+            for name in files:
+                path = os.path.join(root, name)
+                self.logger.debug(f'Removing file from tmp dir: {path}')
+                os.remove(path)
+            for name in dirs:
+                path = os.path.join(root, name)
+                self.logger.debug(f'Removing directory from tmp dir: {path}')
+                os.rmdir(path)
```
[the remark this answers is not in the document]
TonyRice: Mainly for debugging, but either way I can do that too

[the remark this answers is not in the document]
wilzbach: If the user created many files, this manual walk might take a few milliseconds or even longer.
With a rmdir, the OS can - depending on the underlying file system - simply free up the one dir node. In either case, it will be vastly faster.

tl;dr: you want to use https://docs.python.org/3/library/shutil.html#shutil.rmtree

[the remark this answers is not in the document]
TonyRice: Ah, that definitely makes more sense! I'll update

tests/unit/App.py
```diff
@@ -212,6 +214,37 @@ def test_app_init(magic, config, logger, env, always_pull_images):
...
 
+def test_app_get_tmp_dir(app):
+    assert app.get_tmp_dir() == '/tmp/story.app_uuid'
+
+
+def test_app_create_tmp_dir(patch, app):
+    patch.object(pathlib, 'Path')
+    patch.object(app, 'get_tmp_dir')
+
+    # Yes, called twice to ensure the dir is created just once.
```
wilzbach:
```suggestion
    # create_tmp_dir is called twice to ensure the dir is created just once.
```

You could also add this in the test comment or test name.

storyruntime/App.py
```diff
@@ -96,6 +100,26 @@ def image_pull_policy(self):
...
 
+    def create_tmp_dir(self):
+        if self._tmp_dir_created:
+            return
+
+        self._tmp_dir_created = True
+
+        path = self.get_tmp_dir()
+        pathlib.Path(path).mkdir(parents=True, mode=0o700, exist_ok=True)
+        self.logger.debug(f'Created tmp dir {path} (on-demand)')
```
wilzbach: Maybe log before, s.t. you know what failed when it does?

storyruntime/App.py
```diff
@@ -96,6 +100,26 @@ def image_pull_policy(self):
...
         await self.run_stories()
 
+    def create_tmp_dir(self):
+        if self._tmp_dir_created:
+            return
+
+        self._tmp_dir_created = True
+
+        path = self.get_tmp_dir()
+        pathlib.Path(path).mkdir(parents=True, mode=0o700, exist_ok=True)
```
wilzbach: This still might fail.

storyruntime/App.py
```diff
@@ -96,6 +100,29 @@ def image_pull_policy(self):
...
+            return
+
+        self._tmp_dir_created = True
+
+        path = self.get_tmp_dir()
+        self.logger.debug(f'Creating tmp dir {path} (on-demand)')
+        try:
+            pathlib.Path(path).mkdir(parents=True, mode=0o700, exist_ok=True)
+        except BaseException as e:
+            self.logger.error(f'Failed to create tmp dir {path}', e)
```
wilzbach: Shouldn't we let the app crash if this fails?

TonyRice: I went ahead and updated it to re-throw the exception, but I'm not entirely sure how we would test something without a basic unit test.

Issue #479, opened 2019-08-22
Sanitize file service errors


TonyRice:
Currently, file system errors will leak the paths on the system fs. Ideally, we should remove that from the error message.

PR #482, opened 2019-08-22
Sanitizing FileNotFoundError in file service

TonyRice:


wilzbach:
As you're working on https://github.com/storyscript/runtime/issues/479, you might want to fix the root problem as well: Pokemon caught exceptions being printed to the user.

TonyRice:
> As you're working on #479, you might want to fix the root problem as well: Pokemon caught exceptions being printed to the user.

Good idea, I'm going to go over any services and work on handling more specific exceptions. Is there an existing issue with that?

wilzbach:
> Is there an existing issue with that?

Now there's one: https://github.com/storyscript/runtime/issues/484

TonyRice:
#483 already has this merged into it since that's where the most sanitizing would happen.

storyruntime/processing/internal/File.py
```diff
@@ -24,6 +24,10 @@ def safe_path(story, path):
     return f'{story.get_tmp_dir()}{os.fspath(path)}'
 
 
+def clean_path(story, path):
+    return path.replace(story.get_tmp_dir(), '')
```
wilzbach: Use `relative_to` instead: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to

It will throw a proper error when it fails to do so + it will handle OS-specific path details.

PR #485, opened 2019-08-23
Http service enhancements

TonyRice:
This adds support for http request which returns an object with the requests response headers. It allows for http 500 errors to be handled, etc. Instead of doing this in fetch, we do it in a separate method to avoid breaking things

### DIFF
--- a/storyruntime/processing/internal/Http.py
+++ b/storyruntime/processing/internal/Http.py
@@ -73,5 +73,67 @@ async def http_post(story, line, resolved_args):
     return response.body.decode(charset)
 
 
+@Decorators.create_service(name='http', command='request', arguments={
+    'url': {'type': 'string'},
+    'headers': {'type': 'map'},
+    'body': {'type': 'string'},
+    'method': {'type': 'string'}
+}, output_type='any')
+async def http_request(story, line, resolved_args):
+    method = resolved_args.get('method', 'get') or 'get'
+    http_client = AsyncHTTPClient()
+    kwargs = {'method': method.upper(), 'ca_certs': certifi.where()}
+
+    headers = resolved_args.get('headers') or {}
+    if headers.get('User-Agent') is None:
+        headers['User-Agent'] = 'Storyscript/1.0-beta'
+
+    kwargs['headers'] = headers
+
+    if resolved_args.get('body'):
+        kwargs['body'] = resolved_args['body']
+        if isinstance(kwargs['body'], dict):
+            kwargs['body'] = json.dumps(kwargs['body'])
+
+    response = await HttpUtils.fetch_with_retry(3, story.logger,
+                                                resolved_args['url'],
+                                                http_client, kwargs)
+
+    resp_headers = {}
+
+    for header in response.headers:
+        resp_headers[header] = response.headers[header]
+
+    charset = 'utf-8'
+
+    content_type = response.headers.get('Content-Type')
+    if content_type is not None and \
+            'charset' in content_type:
+        parsed = cgi.parse_header(content_type)
+
+        if len(parsed) > 1:
+            charset = parsed[1].get('charset')
+
+    body = response.body.decode(charset)
+
+    if 'application/json' in response.headers.get('Content-Type'):
+        try:
+            body = json.loads(body)
+        except json.decoder.JSONDecodeError:
+            story.logger.warn(
+                f'Failed to parse response as JSON, '
+                f'although application/json was specified! '
+                f'response={response.body.decode(charset)}')
+
+    # currently returning a json obj, rather
+    # than something like a namedtuple, for simplicity
+    return {
+        'status_code': response.code,
+        'status': response.reason,
+        'headers': resp_headers,
+        'body': body
+    }
+
+
 def init():
     pass
--- a/tests/unit/processing/internal/Http.py
+++ b/tests/unit/processing/internal/Http.py
@@ -102,5 +102,97 @@ async def test_service_http_fetch(patch, story, line, json_response,
             assert result == fetch_mock.body.decode(charset)
 
 
+@mark.parametrize('method', [['post', 200], ['get', 201],
+                             ['post', 500], ['get', 500]])
+@mark.parametrize('json_response', [True, False])
+@mark.parametrize('user_agent', [None, 'super_cool_agent'])
+@mark.parametrize('body', [True, False])
+@mark.parametrize('json_request', [True, False, False])
+@mark.parametrize('charset', ['utf-8', 'utf-8', 'utf-16'])
+@mark.asyncio
+async def test_service_http_request(patch, story, line, json_response,
+                                    user_agent, body, json_request,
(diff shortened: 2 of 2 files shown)


wilzbach:
>  we do it in a separate method to avoid breaking things

We can still break things. We're not in v1 yet.

> This adds support for http request which returns an object with the requests response headers. It allows for http 500 errors to be handled, etc. 

I'm not sure we should add a separate method for this. How about
a) change the return type of `fetch`
b) add an attribute to `fetch` to return the full response body

TonyRice:
Note: tests are failing because https://github.com/storyscript/http needs an update

judepereira:
@TonyRice @wilzbach lots of stories will break. Can't we do this any other way?

TonyRice:
> @TonyRice @wilzbach lots of stories will break. Can't we do this any other way?

I changed it up a bit so we can avoid breaking any stories

TonyRice:
Might have to push up some changes

TonyRice:
See https://github.com/storyscript/http/pull/50 @judepereira

TonyRice:
@judepereira can you check this pr out?

storyruntime/processing/internal/Http.py
```diff
@@ -33,10 +36,39 @@
...
+
+        assert re.match(re.compile(
+            r'^(?:http)s?://'
+            r'(?:(?:[A-Z0-9]'
+            r'(?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+'
+            r'(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|'
+            r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'
+            r'(?::\d+)?'
+            r'(?:/?|[/?]\S+)$', re.IGNORECASE),
+            url) is not None
```
wilzbach: Can't we use `urlparse` and enforce that it has a `scheme` host. I don't want us to end up with this monster regex:

https://stackoverflow.com/a/835527/3026245

storyruntime/processing/internal/Http.py
```diff
@@ -33,10 +36,39 @@
...
+
     response = await HttpUtils.fetch_with_retry(3, story.logger,
-                                                resolved_args['url'],
+                                                url,
                                                 http_client, kwargs)
 
+    resp_headers = {}
+
+    for header in response.headers:
+        resp_headers[header] = response.headers[header]
```
wilzbach: resp_headers.update(response.headers)

or just do: `resp_headers = response.headers.copy()` above

storyruntime/processing/internal/Http.py
```diff
@@ -47,30 +79,55 @@
...
+    fetch_body = not resolved_args.get('full', False)
+    # decode_body = False can return raw binary
+    decode_body = resolved_args.get('decode', True)
 
+    # Attempt to read the response body.
+    response_body = response.body
+
+    # if we are calling fetch body, the default functionality
+    # will fail without a valid status code
+    if fetch_body and int(response.code / 100) != 2:
```
wilzbach:
```suggestion
    if fetch_body and response.code // 100 != 2:
```

storyruntime/processing/internal/Http.py
```diff
@@ -47,30 +79,55 @@
...
 
-    if int(response.code / 100) != 2:
-        # Attempt to read the response body.
-        response_body = None
-        try:
-            response_body = response.body.decode(charset)
-        except UnicodeDecodeError:
-            pass
+    # by default we will fetch the body
+    fetch_body = not resolved_args.get('full', False)
```
wilzbach: Sorry for the nit-picking, but `full` is a pretty bad name as it could refer to the body. You have already fetched the body and return it even with `full: false`.

storyruntime/processing/internal/Http.py
```diff
@@ -47,30 +79,55 @@
...
+        return response_body
+
+    # currently returning a json obj, rather
+    # than something like a namedtuple, for simplicity
+    return {
+        'status_code': response.code,
+        'status': response.reason,
+        'headers': resp_headers,
+        'body': response_body
+    }
```
wilzbach: Uhm this is bad because now Storyscript has no typing information because you have to return `any`.

judepereira: +1 for that. I don't like this behaviour of one action returning two possible structures. We should have a different action for this.

judepereira: Can we start this process from reviewing the proposed changes to `http`'s microservice.yaml?

storyruntime/processing/internal/Http.py
TonyRice: Yeah I'll get that in there, that is what I originally used :D

storyruntime/processing/internal/Http.py
TonyRice: No worries at all! I think ```raw``` makes more sense too so I'll do that

storyruntime/processing/internal/Http.py
TonyRice: I personally think we should have two types of methods, ```fetch``` which will return the response body, and ```request``` which will return the full request.

storyruntime/processing/internal/Http.py
```diff
@@ -12,17 +13,19 @@
...
+    'url': {'type': 'string', 'required': True},
+    'headers': {'type': 'map', 'required': False},
+    'body': {'type': 'string', 'required': False},
+    'method': {'type': 'string', 'required': False},
+    'full': {'type': 'boolean', 'required': False},
+    'decode': {'type': 'boolean', 'required': False}
 }, output_type='any')
 async def http_post(story, line, resolved_args):
-    method = resolved_args.get('method', 'get') or 'get'
+    method = resolved_args.get('method', 'GET')
```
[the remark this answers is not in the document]
TonyRice: A previous update in the past I made it possible to avoid this by not including any values that are not required. So if it's required, it will result in None. In this case, the method isn't required so it will hit the default 'GET'

See Services.execute_internal https://github.com/storyscript/runtime/blob/master/storyruntime/processing/Services.py#L156

storyruntime/processing/internal/Http.py
```diff
@@ -47,30 +79,55 @@
...
-            return json.loads(response.body.decode(charset))
-        except json.decoder.JSONDecodeError:
+            response_body = response.body.decode(charset)
+        except UnicodeDecodeError:
             story.logger.warn(
-                f'Failed to parse response as JSON, '
-                f'although application/json was specified! '
+                f'Failed to automatically decode the response body! '
                 f'response={response.body.decode(charset)}')
+            pass
```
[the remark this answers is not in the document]
TonyRice: That pass was not meant to be there. The warning only happens because the original response_body is stored as it's the original state. If decoding fails, a warning is given and the raw response body is given. The user can do what they wish with this. The other route we can take is throw an error unless they disable decoding

PR #487, opened 2019-08-27
Updates for immutable List mutations

TonyRice:
Not sure if we should just go with a shallow copy for this, or if a deep copy is needed for the list.


wilzbach:
Deep copy is required. Storyscript doesn't have head-const immutability, but full transitive immutability.

storyruntime/processing/mutations/ListMutations.py
```diff
@@ -19,24 +19,32 @@ def length(cls, mutation, value, story, line, operator):
     @classmethod
     def append(cls, mutation, value, story, line, operator):
         item = story.argument_by_name(mutation, 'item')
-        return value.append(item)
+        copy = value.copy()
```
wilzbach:
```suggestion
        copy = value.deepcopy()
```

judepereira:
Deep copying from the root call at https://github.com/storyscript/runtime/blob/6d50276314791d8e066d4b1ce505483f1668d37d/storyruntime/processing/Mutations.py#L13. Will open up a new PR.

TonyRice:
> Deep copying from the root call at
> 
> https://github.com/storyscript/runtime/blob/6d50276314791d8e066d4b1ce505483f1668d37d/storyruntime/processing/Mutations.py#L13
> 
> . Will open up a new PR.

We should probably communicate with each other directly when we're making decisions to invalidate someone's code without any clear reason. For example: "I think this solution might also work because of XXX"

The solution in: https://github.com/storyscript/runtime/pull/547 definitely works great, btw.

judepereira:
Got it, will do. Thanks @TonyRice!

PR #501, opened 2019-09-05
Increasing request timeouts in services

TonyRice:


TonyRice:
This helps solve some issues with certain services taking too long to respond 20+ seconds.

storyruntime/processing/Services.py
```diff
@@ -483,9 +483,15 @@ def smart_insert(cls, story, line, command_conf: dict, key: str, value,
...
         method = command_conf['http'].get('method', 'post')
+
+        # set the default request timeout
+        # to 2 minutes, and set the retry
+        # timeout to 1s
         kwargs = {
             'method': method.upper(),
-            'headers': header_params
+            'headers': header_params,
+            'request_timeout': 60000 * 2,
```
adnrs96: This should be 60*2
I believe asyncio.sleep accepts seconds
https://docs.python.org/3/library/asyncio-task.html#sleeping

TonyRice: it's a bit confusing to look at but, request_timeout is passed to AsyncHTTPClient which requires ms, the retry_timeout is what's passed to sleep

adnrs96: ahh, I misread a bit

PR #505, opened 2019-09-10
updating runtime with the "forward" option

TonyRice:
Currently the OMG cli validates against "forward", rather
than forwards as listed in the docs. This is a "hotfix", to
help existing services utilize the forward and forwards option
until the spec is updated.


TonyRice:
> I would recommend adding a comment that explains that this is only temporary. Additionally, you might want to add a helper function for this dictionary operation as this is quite a bit of repetitive code.

I do have a few changes I also need to push up. I'll go ahead and shrink this down and add a good comment.

storyruntime/Apps.py
```diff
@@ -205,7 +205,8 @@ def get(cls, app_id: str):
         services = {}
         all_services = stories.get('services', [])
 
-        expose = asyncy_yaml.get(KEY_FORWARDS, asyncy_yaml.get(KEY_EXPOSE, {}))
+        expose = Dict.default(asyncy_yaml, EXPOSE_KEYS, [])
```
wilzbach: The default changed from empty dict to empty array.

storyruntime/utils/Dict.py
```diff
@@ -19,6 +21,26 @@ def set(_dict, keys, output):
 
             _cur[Resolver.resolve(last, _dict)] = output
 
+    @staticmethod
+    def default(item: dict, keys: [], default=None):
+        """
+        This simply will allow you to search for a key, and
```
wilzbach: Simply is colloquial. Also, documentation is typically written in the present tense.

storyruntime/utils/Dict.py
```diff
@@ -19,6 +21,26 @@ def set(_dict, keys, output):
 
             _cur[Resolver.resolve(last, _dict)] = output
 
+    @staticmethod
+    def default(item: dict, keys: [], default=None):
```
wilzbach: The name is quite generic, maybe sth. like `multi_default` to show future readers the intent without needing them to read the docs.

storyruntime/utils/Dict.py
```diff
@@ -32,11 +54,23 @@ def find(root, path, default_value=None):
...
+
+        if not isinstance(path, list):
+            paths.append(path)
+        else:
+            paths = path
+
+        for p in paths:
+            assert isinstance(p, str)
+            tokens = p.split('.')
+            r = copy.deepcopy(root)
```
wilzbach: Why do we need to do this here?

storyruntime/App.py
```diff
@@ -135,10 +135,21 @@ def get_tmp_dir(self):
         self.logger.info(f'Exposing service {e.service}/'
                          f'{e.service_forward_name} '
                          f'on {e.http_path}')
-        conf = Dict.find(self.services,
-                         f'{e.service}'
-                         f'.{ServiceConstants.config}'
-                         f'.expose.{e.service_forward_name}')
+
+        # this is a temporary hotfix for the current
+        # forward/forwards issue in the omg.
```
wilzbach: Maybe link to the relevant issue if it has a ticket.

storyruntime/utils/Dict.py
```diff
@@ -26,17 +26,29 @@ def find(root, path, default_value=None):
...
-            root = root.get(token)
-            if root is None:
-                return default_value
+        paths = []
 
-        return root
+        if not isinstance(path, list):
+            paths.append(path)
+        else:
+            paths = path
```
wilzbach: Could be 1 line instead of 5, e.g.

```py
paths = path if isinstance(path, list) else [path]
```

Though it might be better to just accept a list for better error messages and typing.

storyruntime/utils/Dict.py
```diff
@@ -26,17 +26,29 @@ def find(root, path, default_value=None):
...
+
+        for p in paths:
+            assert isinstance(p, str)
+            tokens = p.split('.')
+            r = root.copy()
```
wilzbach: Why is this necessary?

storyruntime/utils/Dict.py
```diff
@@ -26,17 +26,29 @@ def find(root, path, default_value=None):
...
+            for token in tokens:
+                r = r.get(token)
+                if r is None:
+                    break
```
wilzbach: Technically a dictionary entry could be `None` and thus the default value would be wrongfully returned.

```py
a = {'a': None}
Dict.find(a, 'a', 2) // None, not 2
```

judepereira:
@TonyRice ? The `forwards` is pretty new in the spec - why not just fix the validator instead of introducing additional code in the runtime?

TonyRice:
> @TonyRice ? The `forwards` is pretty new in the spec - why not just fix the validator instead of introducing additional code in the runtime?

There was a discussion on Slack, currently the OMG outlines forward, and forwards. The validator validates against forward. So it needs to be fixed there. This was a hotfix. I have some apps locally deployed that I would love to deploy in the Storyscript cloud, but I can't do so until this is fixed.

judepereira:
Since they’re local apps, let’s fix them instead (since they’re easy to fix)? Adding this to the runtime is really unnecessary. 


Cheers,
Jude Pereira

TonyRice:
> Since they’re local apps, let’s fix them instead (since they’re easy to fix)? Adding this to the runtime is really unnecessary. Cheers, Jude Pereira
> […](#)
> Op 19 sep. 2019 om 20:37 heeft Tony Rice ***@***.***> het volgende geschreven: ﻿ @TonyRice ? The forwards is pretty new in the spec - why not just fix the validator instead of introducing additional code in the runtime? There was a discussion on Slack, currently the OMG outlines forward, and forwards. The validator validates against forward. So it needs to be fixed there. This was a hotfix. I have some apps locally deployed that I would love to deploy in the Storyscript cloud, but I can't do so until this is fixed. — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or mute the thread.

Got it, we just need to get the docs fixed and the OMG cli updated then.

judepereira:
Thank you :)
Less code in the runtime is easier to manage.


Cheers,
Jude Pereira

Building OSS @ Asyncy, StoryScript, and OMG.

Issue #333, opened 2019-06-12
Global immutable variables


wilzbach:
1) There's no need to copy `app` - it is immutable

https://github.com/storyscript/platform-engine/blob/ef77bd81cdc7dbaad9e45c58392924d739de29d0/asyncy/Stories.py#L259-L261

2) A user must be able to define immutable global variables.

Issue #328, opened 2019-06-12
Function context should be aware of globals


wilzbach:
Currently the function context only includes the function arguments.

PR #507, opened 2019-09-12
Add global context and stacked contexts

anukul:
Closes https://github.com/storyscript/runtime/issues/329, also closes https://github.com/storyscript/runtime/issues/328, also closes https://github.com/storyscript/runtime/issues/333

Deployed to staging, please feel free to run stories and check.


anukul:
Will update tests in some time

storyruntime/processing/Lexicon.py
```diff
@@ -104,7 +105,8 @@ class Lexicon:
                 elif method == 'continue':
                     return await Lexicon.continue_(logger, story, line)
                 elif method == 'while':
-                    return await Lexicon.while_(logger, story, line)
+                    with story.new_context():
+                        return await Lexicon.while_(logger, story, line)
                 elif method == 'try':
                     return await Lexicon.try_catch(logger, story, line)
```
wilzbach: try/catch could also create a new scope.

storyruntime/processing/Lexicon.py
```diff
@@ -175,12 +187,12 @@ class Lexicon:
         This will setup a new context for the
         function block to be executed, and will return the output (if any).
         """
-        current_context = story.context
+        old_context = story._context.copy()
```
wilzbach: why do we need to copy here?

storyruntime/Story.py
```diff
@@ -48,6 +48,41 @@ def new_frame(self, line_number: str):
...
+        Returns the global context for the story
+        """
+        return self.app.story_global_context.setdefault(self.name, {})
+
+    def resolve_context(self, variable):
+        """
+        Used by set_variable to determine the context for a given variable
+        """
+        global_context = self.global_context()
+        for ctx in self._context + [global_context]:
```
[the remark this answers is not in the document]
wilzbach: Also you want to go in reversed order:

```py
chain(reversed(self._context), (global_context,))
```

storyruntime/processing/Lexicon.py
```diff
@@ -104,7 +105,8 @@ class Lexicon:
                 elif method == 'continue':
                     return await Lexicon.continue_(logger, story, line)
                 elif method == 'while':
```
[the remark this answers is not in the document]
anukul: Yeah the `when` block is executed with a new context by StoryEventHandler

storyruntime/processing/Lexicon.py
anukul: The context is modified during function execution

storyruntime/Story.py
```diff
@@ -48,6 +48,41 @@ def new_frame(self, line_number: str):
...
+            if variable in ctx.keys():
+                return ctx
+        # variable not found in context
+        return self._context[-1] if len(self._context) > 0 else global_context
+
+    def get_context(self):
+        """
+        Returns the current context, i.e. all variables in scope
+        """
+        context = self.global_context().copy()
```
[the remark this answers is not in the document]
anukul: 1. This was previously invoked to get a dict containing all variables in scope, which was used to resolve paths that could be global.
2. `copy` because we don't want to update the global context and need to create a new dict (see next few lines).

Although this function will largely be useless after the last commit that involves changes in the `Resolver` so it uses `Story.resolve_context` to resolve paths.

storyruntime/Story.py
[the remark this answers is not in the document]
anukul: Is reversed order important?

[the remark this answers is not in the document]
wilzbach: You want to start with the most-nested context first.

storyruntime/Story.py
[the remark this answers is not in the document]
wilzbach: So what you want is sth. like:

```py
context = {}
context_items = chain.from_iterable(d.items() for d in reversed(self._context))
for k, v in chain(context_items, global_context.items()):
  if k not in context:
     context[k] = v
```

This doesn't perform unnecessary map overwrites.

storyruntime/processing/Lexicon.py
wilzbach: Don't you create a new one with `context = story.context_for_function_call(line, function_line)` and then overwrite it with `story.set_context(context)`?

anukul: Ah python variables got me again :P
Thanks!

judepereira:
I'm taking over this PR after some discussion with Anukul, since he's currently busy with the playground.

storyruntime/processing/Lexicon.py
```diff
@@ -85,7 +85,8 @@ class Lexicon:
                 if method == 'if' or method == 'else' or method == 'elif':
                     return await Lexicon.if_condition(logger, story, line)
```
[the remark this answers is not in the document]
anukul: Should we address this later, once the related changes in the compiler are released?

storyruntime/Story.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, app, story_name, logger):
         self.entrypoint = app.stories[story_name]['entrypoint']
         self.results = {}
         self.environment = None
-        self.context = None
+        self._context = []
```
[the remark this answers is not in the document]
wilzbach: How was this resolved? You can even see that this style is used in this codebase _two_ lines above.

storyruntime/Story.py
```diff
@@ -49,6 +49,44 @@ def new_frame(self, line_number: str):
...
+        """
+        self._context.append({})
+        yield
+        self._context.pop()
+
+    def global_context(self):
+        """
+        Returns the global context for the story
+        """
+        return self.app.story_global_context.setdefault(self.name, {})
```
wilzbach: http://wiki.c2.com/?LawOfDemeter

This logic should be defined with global_context.

storyruntime/Story.py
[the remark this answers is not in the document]
anukul: Oh, looks like I completely overlooked it. Sorry about that.

storyruntime/App.py
```diff
@@ -51,6 +51,7 @@ class App:
 
     def __init__(self, app_data: AppData):
         self._subscriptions = {}
+        self.story_global_context = {}
```
wilzbach: Why is this defined within app?
The top-level context of a story is _not_ shared with other stories.

storyruntime/Story.py
```diff
@@ -49,6 +49,44 @@ def new_frame(self, line_number: str):
...
+        Used by set_variable to determine the context for a given variable
+        """
+        global_context = self.global_context()
+        for ctx in chain(reversed(self._context), (global_context,)):
+            if variable in ctx:
+                return ctx
+        # variable not found in context
+        return self._context[-1] if len(self._context) > 0 else global_context
+
+    def get_context(self):
```
wilzbach: As mentioned before, this isn't a getter. It does a lot more work that that. It should be named accordingly, e.g.

```suggestion
    def build_combined_context(self):
```

storyruntime/processing/Lexicon.py
```diff
@@ -104,9 +105,11 @@ class Lexicon:
                 elif method == 'continue':
                     return await Lexicon.continue_(logger, story, line)
                 elif method == 'while':
-                    return await Lexicon.while_(logger, story, line)
+                    with story.new_context():
+                        return await Lexicon.while_(logger, story, line)
                 elif method == 'try':
-                    return await Lexicon.try_catch(logger, story, line)
+                    with story.new_context():
+                        return await Lexicon.try_catch(logger, story, line)
```
wilzbach: The following is valid Storyscript and would be broken with this line:

```
try
   a = myService
catch
   a = 0

b = a + 1 # OK
```

https://github.com/storyscript/storyscript/issues/1227 should help to fix this, but until then maybe simply not create a new scope here?

storyruntime/processing/Lexicon.py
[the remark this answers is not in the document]
wilzbach: It will be at least two weeks until we will release 0.26 (0.25 will be this weekend). In any case, https://github.com/storyscript/storyscript/issues/1227 is required as well, so I think it's fine to leave as is until it's fixed in the compiler.

storyruntime/processing/Lexicon.py
```diff
@@ -141,12 +144,22 @@ class Lexicon:
...
-                    story.context[ContextConstants.service_event].get('data')
+            story.set_variable(
+                assign={'paths': [ContextConstants.service_output]},
+                output=parent_line['output'][0]
+            )
+
+            if story.resolve({
+                '$OBJECT': 'path',
+                'paths': [ContextConstants.service_event]
+            }) is not None:
```
wilzbach: Better save the resolved context here to avoid searching for it again four lines below.

storyruntime/App.py
anukul: A `when` block creates a new `Story` instance. I'll rename this to `story_global_contexts`

storyruntime/utils/Dict.py
```diff
@@ -10,14 +10,13 @@ def set(_dict, keys, output):
             _dict[keys[0]] = output
         else:
             _cur = _dict
-            last = keys[-1]
-            for key in keys[:-1]:
+            intermediate, last = keys[:-1], keys[-1]
```
wilzbach: Don't create tuples because you don't want to use two lines.

storyruntime/processing/Lexicon.py
```diff
@@ -141,12 +144,22 @@ class Lexicon:
...
+                story.set_variable(
+                    assign={'paths': parent_line['output']},
+                    output=story.resolve({
+                        '$OBJECT': 'path',
+                        'paths': [ContextConstants.service_event]
+                    }).get('data')
+                )
```
wilzbach: With the above change (assuming the object is set to `service_event_path`) this could be sth. more readable like:

```
output_name = parent_line['output']
context[output_name] = context[service_event_path]
```

storyruntime/utils/Dict.py
anukul: I used a tuple here because it fully unpacked the array, the number of lines didn't change.

So, do I change it back?

storyruntime/Story.py
anukul: Not sure how to resolve this.. do I just make a local variable for `self.app.story_global_context` and run `setdefault` on that?

storyruntime/processing/Lexicon.py
anukul: Can you recheck? I just pushed a commit with the last change you suggested, and I'd prefer to use `set_variable` for consistency.

storyruntime/utils/Dict.py
wilzbach: No, you don't unpack it, you do `(keys[:-1], keys[-1])` and then destructure this which is equivalent to:

```py
intermediate = keys[:-1]
last = keys[-1]
```

What you wanted to write was this:

```py
*intermediate, last = keys
```

anukul: Neat, thanks!

tests/integration/processing/Lexicon.py
```diff
@@ -1964,3 +1951,61 @@
...
+                          '    total += x',
+        cases=[
+            Case(assertion=ContextAssertion(key='total',
+                                                expected=6)),
+            Case(assertion=ContextAssertion(key='prefix',
+                                                expected=None)),
+            Case(assertion=ContextAssertion(key='name',
+                                                expected='nobody')),
+            Case(assertion=ContextAssertion(key='greeting',
+                                                expected='Hello user'))
```
wilzbach: This isn't the correct behavior. Functions CANNOT mess with their globals - they are readonly.

storyruntime/Story.py
```diff
@@ -49,6 +49,48 @@ def new_frame(self, line_number: str):
...
+        """
+        self._contexts.append({})
+        yield
+        self._contexts.pop()
+
+    def global_context(self):
+        """
+        Returns the global context for the story
+        """
+        return self.app.story_global_contexts.setdefault(self.name, {})
```
wilzbach: My point with the Law of Demeter was that this behavior should be created wherever global context is defined as this then helps to separate concern.
For example, the global context should be a only readonly dict after story initialization (and not a mutable dictionary).

tests/integration/processing/Lexicon.py
anukul: Tests run with `globals: True`, so I ignored the read-only globals requirement as we do in most other tests.

Do I change this?

anukul: I was trying to emulate the behavior where a function is defined in a nested scope and modifies those variables, but now that I actually tried it, I get https://github.com/storyscript/storyscript/issues/1228

wilzbach: > Although now I don't think that would work at all in the runtime, even if this compiled.

Yes, it's a valid error (all ICEs are important!), though the compiler will just throw an error once the issue is fixed. We don't support closures as of now.

> so I ignored the read-only globals requirement as we do in most other tests.

The only reason for running the tests with `globals: True` is because it's easier to ignore the error message when writing tests.

> Do I change this?

Well, the thing is that the compiler would prevent the above test case:

```coffee
a = 1
function foo
 a = 2
```

errors with:

<details>

```
Error: syntax error in foo.story at line 3, column 2

3|     a = 2
       ^^^

E0127: `a` is readonly and can not be assigned to.
```

</details>

As we have to revisit a lot of logic if/when we ever have mutable globals, I guess you could keep the test case, but maybe add a comment for future readers?

anukul: Cool, in that case I removed it because my intention was to check for a closure, and that neither works in the compiler nor in the runtime.

storyruntime/Story.py
anukul: Cool, updated for the first part.
I'm not sure how to make it readonly in the way we want (no adding / deleting keys).

wilzbach:
@anukul re const dictionary:

```py
class ConstDict:
    """
    Like a dict, but with constant attributes
    """

    def __init__(self, data):
        self._data = data

    def __getattr__(self, attr):
        return self._data[attr]

    def __getitem__(self, item):
        return self._data[item]

    def keys(self):
        return self._data.keys()
```

anukul:
@wilzbach thanks. do we really need this? seems overkill to me

wilzbach:
What's the "overkill" in ten lines which ensure that no one ever mess with the globals?

anukul:
A custom dict class for this purpose.

Anyway, I'll go ahead and it. :P

tests/unit/Story.py
```diff
@@ -89,6 +90,49 @@ def test_story_line(magic, story):
...
+        story.name: {'a': 1, 'b': 2, 'c': 3}
+    })
+    story._contexts = [
+        {'d': 4, 'e': 5, 'f': 6},
+        {'g': 7, 'h': 8, 'i': 9}
+    ]
+    assert story.build_combined_context() == {
+        **app.story_global_contexts[story.name],
+        **story._contexts[0],
+        **story._contexts[1],
```
wilzbach: Better to write out the full dict here to make sure the function actually does behave as desired,

tests/unit/utils/ConstDict.py
```diff
@@ -0,0 +1,32 @@
...
+    assert const_dict.c == normal_dict['c']
+    # getitem
+    for key, value in normal_dict.items():
+        assert const_dict[key] == value
+    # keys
+    assert normal_dict.keys() == const_dict.keys()
+    # setattr
+    with pytest.raises(Exception):
+        const_dict.a = 0
+    # setitem
```
wilzbach: All of these should be separate methods.

storyruntime/App.py
```diff
@@ -82,6 +83,10 @@ def __init__(self, app_data: AppData):
             'hostname': f'{self.app_dns}.{self.config.APP_DOMAIN}',
             'version': self.version
         }
+        self.story_global_contexts = ConstDict({
+            story_name: {}
```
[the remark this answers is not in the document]
anukul: So basically no editing in `when` blocks?

wilzbach:
FYI this is only partially covered:

image

anukul:
Is it possible to define the same variable in two scopes?

wilzbach:
> Is it possible to define the same variable in two scopes?

No, currently this is only possible via functions. If you do a `a = 1` but `a` exists somewhere else, Storyscript assumes you overwrite it.

anukul:
Cool, so I've removed the `if` condition to change that partial into a hit.

PR #509, opened 2019-09-12
feat(Lexicon): handle sentinels from executing try catch finally

adnrs96:
This will make sure cases like the following work fine:
```
a = ["foo", "bar"]
foreach a as s
    try
        if s == "bar"
            continue
        when twitter stream tweet track: s as result
            log info msg: result['text']
    finally
        log info msg: "wow"
```

Right now sentinels from within try catch aren't bubbled up.
We should be doing that.

Fixes: #511


adnrs96:
As discussed I have banned the sentinel keywords inside the finally block and left a comment.

This should be ready for a merge after a review. @judepereira can you please review it?

judepereira:
@adnrs96 yes, I'll review it today.

PR #516, opened 2019-09-16
Update README ENV variable typos

matthewhudson:


judepereira:
@wilzbach CI is failing for this. I was investigating why, but then got caught up in a meeting yesterday.

wilzbach:
@judepereira it was an unrelated failure. The failure is due to updated OMG service typing which made newer builds fail.
Anyhow, I think @TonyRice has already fixed this and thus master is green (even with this PR): https://circleci.com/gh/storyscript/runtime/1671

PR #539, opened 2019-09-27
Shorter Exception Messages

TonyRice:
Currently, exceptions can be very large and making it nearly impossible to decode the data. This helps avoid overflowing the logs with large amounts of data.


storyruntime/Exceptions.py
```diff
@@ -15,7 +15,11 @@ def __str__(self):
             return super().__str__()
 
         stack = self.story.get_stack()
-        trace = f'An exception has occurred:\n{self.message}'
+
+        trimmed_msg = f'{self.message[0:125]}...' if \
+            len(self.message) >= 125 else self.message
```
wilzbach: Broken for n == 125.
How about:

```suggestion
        trimmed_msg =  self.message[:MAX_LOG_MESSAGE_LENGTH]
        if len(self.message) > MAX_LOG_MESSAGE_LENGTH:
            trimmed_msg += '...'
```

judepereira: @TonyRice looks good. Let's apply it, add a small test, and merge?

PR #546, opened 2019-09-30
When maps/lists are cast to a string, use their JSON representation.

### LINKED ISSUE #545: When maps are casted to a string, jsonify them by default
Currently, the case to string for maps results in the Python string for a dictionary, which is less friendly as compared to JSON.

judepereira:
Closes #545


storyruntime/utils/TypeResolver.py
```diff
@@ -115,6 +116,10 @@ def check_type_cast(cls, type_exp, item):
             # bytes must be properly decoded
             elif isinstance(item, bytes):
                 return item.decode()
+            elif isinstance(item, dict):
+                return json.dumps(item)
```
TonyRice: We need to ensure that values such as namedtuples, bytes, etc are properly handled.

judepereira: Yes, but that is out of scope for this PR. This one specifically addresses maps and lists.

TonyRice: What I more specifically mean, is that let's say for example we store binary data in a map or something not JSON, then it might cause some issues.

judepereira: Ah yes, let's create another issue for that? It doesn't seem necessary to solve it right now itself, given that it's not easy to throw in data which is not serialisable in a map. Other questions that need to be asked is - should it produce a runtime error? As any strategy we pick to serialise non-JSON serialisable data (like base64 for example) will not be a decent enough solution.

PR #547, opened 2019-09-30
Deep copy

judepereira:


storyruntime/processing/Mutations.py
```diff
@@ -18,7 +20,9 @@ def mutate(cls, mutation, value, story, line):
                 handler = getattr(StringMutations, operator)
             elif isinstance(value, list):
                 handler = getattr(ListMutations, operator)
+                value = copy.deepcopy(value)
```
TonyRice: We should probably add a note to signal developers why we run a deepcopy before running the mutations

storyruntime/processing/Lexicon.py
```diff
@@ -424,6 +424,8 @@ def throw(logger, story, line):
         output = line['output'][0]
 
         try:
+            if _list is None:
```
TonyRice: Need to remove debug message

judepereira: Nice catch. Thanks!

PR #548, opened 2019-10-01
Updating http service to utilize binary instructions for raw data

TonyRice:
See https://github.com/storyscript/http/pull/49


storyruntime/utils/TypeResolver.py
```diff
@@ -113,9 +113,13 @@ def check_type_cast(cls, type_exp, item):
             # bools are always lowercase in SS
             if isinstance(item, bool):
                 return str(item).lower()
-            # bytes must be properly decoded
+            # attempt to decode bytes, otherwise
+            # return it as a string
             elif isinstance(item, bytes):
-                return item.decode()
+                try:
+                    return item.decode('utf-8')
```
TonyRice: @judepereira I was thinking, by default, we attempt to decode the data as a utf-8 string. Most binary data will fail to do this. I'm not sure which way to go on this, but I was thinking by default we could return the data base64 encoded if it cannot be decoded into a utf-8 string.

storyruntime/processing/Services.py
```diff
@@ -295,36 +296,99 @@ def get_command_conf(cls, story, chain):
...
-        # 3. Dump the bytes directly in the response
+        close_connection = True
+
+        next_line = story.line(line.get('next', None))
+
+        # This will go over each line after the current line, and
+        # check if the service is the same, if it is, then we will
+        # continue on to process everything normally. If there are
+        # no other operations in this block for the service, then
+        # we will go ahead and ensure the connection closes.
```
judepereira: Why are we doing this?

storyruntime/processing/Services.py
```diff
@@ -295,36 +296,99 @@ def get_command_conf(cls, story, chain):
...
-            story.app.logger.info('Connection has been closed '
-                                  'for service http implicitly, '
-                                  'as binary data was written to it.')
-            io_loop.add_callback(req.finish)
+            await req.flush()
+
+            # sleep so we have a moment
+            # to effectively send the last raw
+            # instruction
+            await asyncio.sleep(0.1)
```
judepereira: Why? The underlying connection will take care of all this. It's supposed to be a guarantee that flush flushes it to the system buffer.s

storyruntime/processing/Services.py
```diff
@@ -295,36 +296,99 @@ def get_command_conf(cls, story, chain):
...
+            # that we are done sending binary data.
+            # Once this is sent, we can continue to
+            # process other operations like flush
+
+            # the FLUSH operations is the same as
+            # END, except the server will flush
+            # the connection immediately after receiving
+            # the FLUSH instruction
+            if body['data'].get('flush'):
+                req.write(b'\n\nFLUSH\n')
```
judepereira: What happens if the binary data that the runtime is trying to send contains "FLUSH"?

storyruntime/processing/Services.py
TonyRice: The sleep is used to give the connection a chance to send the data off right away. Even if we call flush following a write, the data will be received as one chunk on the other end. I didn't do too much investigating as to why that happens, but this helps solve that. It can probably be removed with a few tweaks on the http service side

storyruntime/processing/Services.py
TonyRice: The raw binary "instructions" are processed as a single chunk avoiding that altogether. Once the binary transport is initiated, all the binary data is sent over until we receive the END or FLUSH command.

storyruntime/processing/Services.py
TonyRice: Originally the implementation automatically closes the connection by adding io_loop.add_callback(req.finish). This will go over any operations after that one to check if there are any other http instructions, if not it'll just go ahead and do the original behaviour and add io_loop.add_callback(req.finish)

storyruntime/processing/Services.py
TonyRice: Let me know if you have any more questions!

storyruntime/processing/Services.py
TonyRice: @judepereira I propose removing io_loop.add_callback(req.finish) and extra code. I don't think there will be a situation where this connection won't automatically close since there are no further instructions for the response.

PR #554, opened 2019-10-02
Pin versions of the test deps to that mentioned in the tox.ini.

judepereira:
The motivation for this is to enable users to simply clone the project, and then be able to run python setup.py test. Will and I explored another option of forcing users to run tox instead of the setuptools test, but were unable to successfully override the test command and print a decent message. Hence, for now, we'll just pin the same versions (since they rarely change anyway).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,10 +23,10 @@ setup(
     version='0.2.0',
     packages=find_packages(),
     tests_require=[
-        'pytest==4.2.0',
-        'pytest-cov==2.6.1',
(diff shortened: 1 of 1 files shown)


jayvdb:
Duplication of test dependencies , especially manually, is not a good solution.

williammartin:
I agree and have created https://github.com/storyscript/runtime/issues/555 to follow up on this. However, right now this is broken for any new contributors, so I'm going to approve and merge. If we shortly have a follow up that resolves this, that would be great also.

PR #556, opened 2019-10-02
chore(test): Sync setup.py and tox test depends

jayvdb:
Also prevent installation of `bench` and tests`, and
focus the coverage on the relevant package only.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,12 +21,14 @@ setup(
     author='Storyscript',
     author_email='noreply@storyscript.io',
     version='0.2.0',
-    packages=find_packages(),
+    packages=find_packages(
+        exclude=('build.*', 'bench', 'bench.*', 'tests', 'tests.*'),
+    ),
     tests_require=[
-        'pytest==4.2.0',
-        'pytest-cov==2.6.1',
-        'pytest-mock==1.10.1',
-        'pytest-asyncio==0.10.0'
+        'pytest',
+        'pytest-cov',
+        'pytest-mock',
+        'pytest-asyncio',
     ],
     setup_requires=['pytest-runner'],
     python_requires='>=3.7',
--- a/tox.ini
+++ b/tox.ini
@@ -1,21 +1,26 @@
 [tox]
+minversion = 1.6
 envlist = py37,pep8
+skip_missing_interpreters = true
 
-
-[testenv:py37]
+[testenv]
 passenv = DOCKER_HOST DOCKER_MACHINE_NAME DOCKER_TLS_VERIFY DOCKER_CERT_PATH
+# These deps are optional here, as `setup.py pytest` will install them anyway
+# if omitted, however tox does it more neatly, and developers may want to try
+# various pins here, and setup.py will not overrule them.
 deps =
-    pytest==3.6.3
-    pytest-cov==2.5.1
-    pytest-mock==1.10.0
-    pytest-asyncio==0.8.0
+    pytest
+    pytest-cov
+    pytest-mock
+    pytest-asyncio
 
 commands =
-    pytest --cov=. --cov-config=.coveragerc --cov-report=term-missing {posargs}
+    python setup.py pytest --addopts="--cov=storyruntime --cov-config=.coveragerc --cov-report=term-missing {posargs}"
     coverage xml
 
-
 [testenv:pep8]
+usedevelop = true
+skip_install = true
 deps =
     flake8==3.5.0
     flake8-quotes==1.0.0


williammartin:
This is still broken for my flow:

```
git checkout -b jayvdb-fix-pytest master
...

git pull https://github.com/jayvdb/platform-engine.git fix-pytest
...

virtualenv venv --python=python3.7 && source venv/bin/activate
...

python setup.py test
...
Traceback (most recent call last):
  File "setup.py", line 58, in <module>
    """
  File "/Users/user/workspace/storyscript/runtime/venv/lib/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest_runner-5.1-py3.7.egg/ptr.py", line 209, in run
    return self.run_tests()
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest_runner-5.1-py3.7.egg/ptr.py", line 220, in run_tests
    result_code = __import__('pytest').main()
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest-4.2.0-py3.7.egg/_pytest/config/__init__.py", line 61, in main
    config = _prepareconfig(args, plugins)
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest-4.2.0-py3.7.egg/_pytest/config/__init__.py", line 182, in _prepareconfig
    config = get_config()
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest-4.2.0-py3.7.egg/_pytest/config/__init__.py", line 156, in get_config
    pluginmanager.import_plugin(spec)
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest-4.2.0-py3.7.egg/_pytest/config/__init__.py", line 530, in import_plugin
    __import__(importspec)
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest-4.2.0-py3.7.egg/_pytest/tmpdir.py", line 25, in <module>
    class TempPathFactory(object):
  File "/Users/user/workspace/storyscript/runtime/.eggs/pytest-4.2.0-py3.7.egg/_pytest/tmpdir.py", line 35, in TempPathFactory
    lambda p: Path(os.path.abspath(six.text_type(p)))
TypeError: attrib() got an unexpected keyword argument 'convert'
```

jayvdb:
@williammartin , you will notice in the tox.ini, I did `setup.py pytest`  , for that reason.  The proper implementation of `setup.py test` for pytest involves overriding the `test` command, as outlined at https://pytest.readthedocs.io/en/2.7.3/goodpractises.html#integration-with-setuptools-test-commands

jayvdb:
I'll add that, and you can test again.

williammartin:
Thanks, I think I'm not sure what is idiomatically expected for Python projects. For example, would a python dev expect to be able to do `python setup.py test`?

jayvdb:
Yes, that has long been the appropriate way to test a Python project, and is still dominant in 'core' packages of the Python ecosystem, however that has waned over the years as invoking `pytest` and `tox` have become the expected ways to test a Python project , and so that extra voodoo to hook up pytest is often omitted.

Some of the other storyscript projets do have the pytest hook in setup.py ; e.g. cli.

Despite the CI failure, which is unrelated, could you test `python setup.py test` now in your environment?  It will be quite curious if it doesnt work, and worth investigating properly because it needs to get solved.

If it does work, I will revert the last change to tox.ini which is just for demonstration purposes.

williammartin:
Yep, will check it out when I get back from lunch. Thanks for the explanation. Is it worth keeping the voodoo or do you think we want to encourage all testing through tox? I'm also curious how to hook this up in intellij test integration, but I'm sure it's a solved problem.

jayvdb:
setup.py should be generated, and be the same for all of our Python projects with substitutions for the minor variations needed; or e.g. poetry and similar generate setup.py for us.  So the answer should be that we use a tool which generates setup.py, and if it doesnt support the `test` command then we accept that decision from the tool and provide alternative instructions for newcomers learning how to get involved in our projects.  The `test` command, or even `pytest` command, is not worth us sticking with `setup.py`.

jayvdb:
To more directly answer, I think we should

1. encourage newcomers use `tox`, to reduce the chance of them breaking their local Python installation, and
2. ensure that `tox -e py38` works correctly, even if it fails because of Python 3.8 stuff. i.e. do not force names our testenvs like 'py37' which means they cant be re-used for other Python versions, and
3. also support running `pytest` directly, which means moving mandatory pytest options from the tox testenv command into metadata that pytest reads -- setup.cfg, pytest.ini or a dedicated settings section in tox.ini .

williammartin:
I see the same behaviour with your latest commit and after creating a new venv to go from scratch:

```
± |jayvdb-fix-pytest ✓| → git lg -n 3
* 64c507d - (HEAD -> jayvdb-fix-pytest) chore(test): Map test command to pytest (51 minutes ago) <John Vandenberg>
* 0ad99ad - chore(test): Sync setup.py and tox test depends (71 minutes ago) <John Vandenberg>
* d6eda69 - Shorter Exception Messages (#539) (3 hours ago) <Tony Rice>
```

I see this both with `python setup.py test` and `python setup.py pytest`

PR #558, opened 2019-10-07
Use config variable for graphql endpoint

anukul:


judepereira:
@anukul LGTM - merge away.

PR #560, opened 2019-10-09
Update to Kubernetes.create_deployment:

TonyRice:
This increases the wait timeout of 1 second to 5 seconds in between each re-try during a deployment

Adds a hard timeout within the while True block (5 minutes)

Adds a check for unavailable resources during deployment.

Currently, the runtime will just wait forever taking up resources.


storyruntime/Kubernetes.py
```diff
@@ -668,13 +668,61 @@ def get_liveness_probe(cls, app, service: str):
 
         # Wait until the deployment is ready.
         app.logger.debug('Waiting for deployment to be ready...')
+
+        fail_count = 0
+
+        # this is used to store when
+        # we started this deployment
+        deploy_start = time.time()
+        default_timeout = 5 * 60
```
wilzbach: Good opportunity to get rid of magic constants, e.g.
- http://wiki.c2.com/?MagicNumber
- http://wiki.c2.com/?ReplaceMagicNumberWithSymbolicConstant

anukul:
@TonyRice Can we add tests for the cases that this PR introduces, seeing as the current test for `create_deployment` includes its existing behavior:

```py
    patch.object(Kubernetes, 'make_k8s_call', new=async_mock(side_effect=[
        _create_response(404),
        _create_response(201),
        _create_response(200, {'status': {'readyReplicas': 0}}),
        _create_response(200, {'status': {'readyReplicas': 0}}),
        _create_response(200, {'status': {'readyReplicas': 1}})
    ]))
```

storyruntime/Kubernetes.py
```diff
@@ -668,13 +668,61 @@ def get_liveness_probe(cls, app, service: str):
...
+
+                    app.logger.debug(
+                        f'The deployment is still not ready, waiting...'
+                    )
+
+                await asyncio.sleep(5)
+                # If we hit this error 25 or more times, then
+                # we can safely assume that we will always
+                # hit this error until the k8 cluster is repaired
+                if fail_count >= 25:
```
wilzbach: http://wiki.c2.com/?MagicNumber

storyruntime/Kubernetes.py
```diff
@@ -659,7 +659,7 @@ def get_liveness_probe(cls, app, service: str):
                 break
 
             app.logger.debug(f'Failed to create deployment, retrying...')
-            await asyncio.sleep(1)
+            await asyncio.sleep(5)
```
judepereira: 5 seconds here is too long. The reason behind keeping the interval as low as possible was to make sure that we don't waste additional time causing an app deployment to be perceived as "slow".

storyruntime/Kubernetes.py
```diff
@@ -668,13 +668,61 @@ def get_liveness_probe(cls, app, service: str):
...
+
+                if not progressing and min_repl:
+                    # we will only hit this if it is not progressing.
+                    fail_count += 1
+
+                    app.logger.debug(
+                        f'The deployment is still not ready, waiting...'
+                    )
+
+                await asyncio.sleep(5)
```
judepereira: Lower the sleep time to 1 second. See reasoning above.

storyruntime/Kubernetes.py
```diff
@@ -668,13 +668,61 @@ def get_liveness_probe(cls, app, service: str):
...
+                # If we hit this error 25 or more times, then
+                # we can safely assume that we will always
+                # hit this error until the k8 cluster is repaired
+                if fail_count >= 25:
+                    raise K8sError(
+                        message=f'Failed to create deployment: '
+                        f'{condition["message"]}'
+                    )
+
             await cls.check_for_image_errors(app, container_name)
```
judepereira: Since you're checking for all errors, do we need to check for image errors explicitly?

TonyRice: This is actually checking for a certain state in the cluster which would normally result in create_deployment waiting forever 'MinimumReplicasUnavailable'. Technically, a cluster could be repaired in the middle of this operation, but we don't want the user waiting forever.

PR #567, opened 2019-10-09
Fix await

judepereira:


judepereira:
@wilzbach @williammartin This PR makes a strong point for E2E tests for the runtime.

PR #574, opened 2019-10-11
[WIP] feat(Regex): add custom Regex class and start entertaining regex flags

adnrs96:
This adds a new class which will handle regex and provide methods to deal with regex and flags.


adnrs96:
@judepereira can you check out the direction I am taking here and let me know what you think

judepereira:
Looks good. Let's add tests and merge it :)

storyruntime/utils/types/RegExp.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
+import re
+
+from storyscript.compiler.semantics.types.Types import RegExpType
+
+
+class RegExp(RegExpType):
```
wilzbach: No. It should inherit from a to-be-created `RuntimeVar`. You want to use encapsulation here.

wilzbach: For reference: https://github.com/storyscript/runtime/issues/327

adnrs96:
@judepereira @wilzbach I updated this PR which is definitely meant to serve as a immediate solution.
In long term we should discuss the regex and how it should behave. And with that my earlier approach of defining a custom type for regex would definitely make sense but for now this should work.

Also storyscript/storyscript#1230 is the issue for dropping some of the flag support in compiler itself.

storyruntime/utils/RegExpUtils.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+import re
+
+
+class RegExpUtils:
+
+    @staticmethod
+    def process_flags(flags):
+        re_flags = 0
```
wilzbach: thinking about we should probably be defensive here and do:

```py
all(x in "ims" for x in flags)
```

adnrs96:
@judepereira Addressed the comment and rebased.

Though @wilzbach I later realized that this is one of those things which would endup error in compilation phase since we won't parse invalid flags in the first place

Issue #585, opened 2019-10-16
Support `foreach list as index, value`


anukul:
Currently only `foreach list as value` is possible.

anukul:
Also `map as key, value`

Issue #575, opened 2019-10-14
Reorganize lexicon integration tests


anukul:
e.g. the `test_mutation` suite tests for stack overflow, internal services (http, file), try / catch etc.

wilzbach:
You might want to use this opportunity to readdress a few issues with the testsuite, e.g.
- https://github.com/storyscript/runtime/issues/268
- https://github.com/storyscript/runtime/issues/529
- https://github.com/storyscript/runtime/issues/416

PR #586, opened 2019-10-16
Support `foreach list as index, value`

anukul:
Closes https://github.com/storyscript/runtime/issues/585

### DIFF
--- a/storyruntime/processing/Lexicon.py
+++ b/storyruntime/processing/Lexicon.py
@@ -440,7 +440,7 @@ class Lexicon:
         for a, b in iterable:
             output = line['output']
             assert 1 <= len(output) <= 2, \
-                f'for loop output must be one or two values'
+                f'for loop output must be 1 or 2 values, found {len(output)}'
             if len(output) == 1:
                 story.set_variable(assign={'paths': output},
                                    output=b if isinstance(data, list) else a)
--- a/storyruntime/processing/Lexicon.py
+++ b/storyruntime/processing/Lexicon.py
@@ -434,12 +434,13 @@ class Lexicon:
         Evaluates a for loop.
         """
         data = story.resolve(line['args'][0], encode=False)
-        assert type(data) in [list, dict]
+        assert type(data) in [list, dict], f'Cannot iterate over {type(data)}'
         iterable = enumerate(data) if isinstance(data, list) else data.items()
 
         for a, b in iterable:
             output = line['output']
-            assert 1 <= len(output) <= 2
+            assert 1 <= len(output) <= 2, \
+                f'for loop output must be one or two values'
             if len(output) == 1:
                 story.set_variable(assign={'paths': output},
                                    output=b if isinstance(data, list) else a)
--- a/storyruntime/processing/Lexicon.py
+++ b/storyruntime/processing/Lexicon.py
@@ -435,8 +435,22 @@ class Lexicon:
         """
         _list = story.resolve(line['args'][0], encode=False)
 
-        for item in _list:
-            story.set_variable(assign={'paths': line['output']}, output=item)
+        for index, element in enumerate(_list):
+            output = line['output']
+            if len(output) == 1:
+                story.set_variable(assign={'paths': output},
+                                   output=element)
+            elif len(output) == 2:
+                story.set_variable(assign={'paths': [output[0]]},
+                                   output=index)
+                story.set_variable(assign={'paths': [output[1]]},
+                                   output=element)
+            else:
+                raise StoryscriptError(
+                    message='for loops can only have one or two outputs',
+                    story=story,
+                    line=line
+                )
 
             result = await Lexicon.execute_block(logger, story, line)
 
--- a/storyruntime/processing/Lexicon.py
+++ b/storyruntime/processing/Lexicon.py
@@ -433,19 +433,19 @@ class Lexicon:
         """
         Evaluates a for loop.
         """
-        _list = story.resolve(line['args'][0], encode=False)
+        data = story.resolve(line['args'][0], encode=False)
+        assert type(data) in [list, dict]
+        iterable = enumerate(data) if isinstance(data, list) else data.items()
 
-        for index, element in enumerate(_list):
+        for a, b in iterable:
             output = line['output']
             assert 1 <= len(output) <= 2
             if len(output) == 1:
                 story.set_variable(assign={'paths': output},
-                                   output=element)
+                                   output=b if isinstance(data, list) else a)
             else:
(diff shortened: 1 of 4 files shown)


storyruntime/processing/Lexicon.py
```diff
@@ -435,8 +435,22 @@ def throw(logger, story, line):
...
+            output = line['output']
+            if len(output) == 1:
+                story.set_variable(assign={'paths': output},
+                                   output=element)
+            elif len(output) == 2:
+                story.set_variable(assign={'paths': [output[0]]},
+                                   output=index)
+                story.set_variable(assign={'paths': [output[1]]},
+                                   output=element)
+            else:
```
anukul: This branch is a miss because the compiler returns an error and we never parse such a syntax tree.

wilzbach: Internal assertions should be an `assert`. 
tl;dr:

```py
            else:
                # output length is enforced by the compiler
                assert len(output) == 2
                story.set_variable(assign={'paths': [output[0]]},
                                   output=index)
                story.set_variable(assign={'paths': [output[1]]},
                                   output=element)
```

tests/integration/processing/Lexicon.py
```diff
@@ -471,6 +471,22 @@
...
+                          '  value_sum += value\n',
+        cases=[
+            Case(
+                assertion=ContextAssertion(key='index_sum', expected=3)
+            ),
+            Case(
+                assertion=ContextAssertion(key='value_sum', expected=27)
+            )
+        ]
+    ),
```
wilzbach: Add a test case for maps

tests/integration/processing/Lexicon.py
```diff
@@ -471,6 +471,38 @@
...
+                          '  value_sum += value\n',
+        cases=[
+            Case(
+                assertion=ContextAssertion(key='key_string', expected='abc')
+            ),
+            Case(
+                assertion=ContextAssertion(key='value_sum', expected=27)
+            )
+        ]
+    ),
```
wilzbach: There's no test case for dict iteration with a single key.

storyruntime/processing/Lexicon.py
```diff
@@ -433,10 +433,20 @@ def throw(logger, story, line):
...
-        for item in _list:
-            story.set_variable(assign={'paths': line['output']}, output=item)
+        data = story.resolve(line['args'][0], encode=False)
+        assert type(data) in [list, dict], f'Cannot iterate over {type(data)}'
+        iterable = enumerate(data) if isinstance(data, list) else data.items()
+
+        for a, b in iterable:
+            output = line['output']
+            assert 1 <= len(output) <= 2, \
+                f'for loop output must be 1 or 2 values, found {len(output)}'
```
wilzbach:
```suggestion
                f'foreach loop output must be 1 or 2 values, found {len(output)}'
```

storyruntime/processing/Lexicon.py
```diff
@@ -433,10 +433,20 @@ def throw(logger, story, line):
         """
         Evaluates a for loop.
```
wilzbach:
```suggestion
        Evaluates a foreach loop.
```

storyruntime/processing/Lexicon.py
```diff
@@ -433,10 +433,20 @@ def throw(logger, story, line):
...
+        assert type(data) in [list, dict], f'Cannot iterate over {type(data)}'
+        iterable = enumerate(data) if isinstance(data, list) else data.items()
+
+        for a, b in iterable:
+            output = line['output']
+            assert 1 <= len(output) <= 2, \
+                f'for loop output must be 1 or 2 values, found {len(output)}'
+            if len(output) == 1:
+                story.set_variable(assign={'paths': output},
+                                   output=b if isinstance(data, list) else a)
```
wilzbach: Can't we split the code into two parts: one for dict and one for array iteration before doing such evil things? They will bite you later on...

tests/integration/processing/Lexicon.py
anukul: [There is](https://github.com/storyscript/runtime/blob/a61eec8b58a7a3b6072aa3787f6e85268440bded/tests/integration/processing/Lexicon.py#L465), but it's all mixed up in `test_mutation`. I'll fix https://github.com/storyscript/runtime/issues/575 soon.

storyruntime/processing/Lexicon.py
anukul: You mean this?
```py
if isinstance(data, list):
    value = b
else: # isinstance(data, dict), but no elif for full coverage
    value = a
story.set_variable(assign={'paths': output}, output=value)
```

anukul: If we write the `for` block twice, we end up duplicating everything else inside :/

wilzbach: 1) Nah you don't need to have everything inside the loop like e.g. this doesn't need to be run on every iteration

```py 
          output = line['output']
          assert 1 <= len(output) <= 2, \
              f'foreach output must be 1 or 2 values, found {len(output)}'
```

2) I don't see how this is a huge duplication

```py
          if isinstance(data, list):
                foreach i, v in enumerate(data):
                     if len(output) == 1:
                         set_output(v)
                    else:
                         set_output(i, v)
```

Also, it's so much more explicit than `a` + `b`.
And if you really want to save line number, you can use arrays and do e.g. `set_output([i] if len(output) == 1 else [i,v])`.

>  we end up duplicating everything else inside

This is good here, because with the duplication you would have noticed that maps with only one output aren't covered.

anukul: > I don't see how this is a huge duplication
>```py
>          if isinstance(data, list):
>                foreach i, v in enumerate(data):
>                     if len(output) == 1:
>                         set_output(v)
>                    else:
>                         set_output(i, v)
>```
The duplication also includes what comes after `set_output`... `execute_block`, `break`, `continue`

> This is good here, because with the duplication you would have noticed that maps with only one output aren't covered.

Sorry I don't understand, what's not covered at the moment?

tests/integration/processing/Lexicon.py
```diff
@@ -471,6 +471,54 @@
             )
         ]
     ),
+    Suite(
+        preparation_lines='l = [8, 9, 10]\n'
+                          'index_sum = 0\n'
+                          'value_sum = 0\n'
+                          'foreach l as index, value\n'
```
judepereira: @anukul If I just specify `foreach l as item`, is item the index or an actual element in the list `l`?

anukul: It is the list element

Issue #512, opened 2019-09-13
try without catch or finally


adnrs96:
Runtime right now assumes that given a `try` statement/block it would always be accompanied by a `catch` or a `finally` block. This is something which isn't checked for or guaranteed by the compiler yet. This means if one would try to compile something like
```
try
    a = 1
    b = 2

c = 4
```

It would succeed in compiling and the runtime would run this particular code.

Issue with the runtime would be since its assuming to have either a `catch` or `finally` block after the `try` block, it is bound to run into errors.

Such a error isn't obvious from the above example but if it was something like

```
try
    a = 1
    b = 2
    throw "wow"

c = 4
```

Now we will see the runtime assume a `catch` block when it doesn't find a `finally` block and try to execute the line `c = 4` as a block.

For cases like below:
```
try
    a = 4
```

The `try` block won't be executed at all, since `try_catch` short circuits if there is no next block to the `try` block.

Given above, we need to either make compiler sure that there isn't a `try` without a `catch` or `finally` or we need to modify the runtime to deal with the lone `try` blocks.

Thoughts on the matter?

judepereira:
Let’s enforce this in the compiler. 


Cheers,
Jude Pereira

wilzbach:
Uhm try without a catch is legit and allowed on purpose. It is equal to just ignoring the exception like you would do e.g. for a service where you don't want your app to crash. For example, sending a log notification or using experimental APIs.

judepereira:
Oh, I see. That makes sense. Yes, we should support a plain try block.


Cheers,
Jude

PR #593, opened 2019-10-16
Adding support for throw without catch or finally

TonyRice:
https://github.com/storyscript/runtime/issues/512


tests/integration/processing/Lexicon.py
```diff
@@ -972,6 +972,15 @@
...
                      key='arr', expected=[1, 42, 42, 3, 4, 4, 5, 5])),
         ]),
+    Suite(
+        preparation_lines='failed = true\n'
+                          'try\n'
+                          '  throw "error"\n',
+        cases=[
+            Case(append='failed = false',
+                 assertion=ContextAssertion(key='failed', expected=False))
+        ]
```
wilzbach: Add this too:

```
    Suite(
        preparation_lines='failed = true\n'
                          'try\n'
                          '  a = 1\n',
        cases=[
            Case(assertion=ContextAssertion(key='failed', expected=True))
        ]
    ),
    Suite(
        preparation_lines='failed = true\n'
                          'try\n'
                          '  throw "error"' 
                          '  failed = false\n'
        cases=[
            Case(assertion=ContextAssertion(key='failed', expected=True))
        ]
    ),

```

storyruntime/processing/Lexicon.py
```diff
@@ -403,7 +403,10 @@ def unless_condition(logger, story, line):
             if LineSentinels.is_sentinel(exec_res):
                 result_sentinel = exec_res
         except StoryscriptError:
-            if next_line['method'] == 'finally':
+            if next_line['method'] != 'finally' and \
+                    next_line['method'] != 'catch':
+                return Lexicon.line_number_or_none(next_line)
```
adnrs96: I don't think we wanna do this over here. Thing is `try` block had an error and I believe if we can't handle it over here we need to raise it upstairs for a higher try catch to handle.
Thoughts?

tests/integration/processing/Lexicon.py
```diff
@@ -1753,6 +1753,33 @@
...
+            Case(assertion=ContextAssertion(key='a', expected=None))
+        ]
+    ),
+    Suite(
+        preparation_lines='failed = true\n'
+                          'try\n'
+                          '  throw "error"\n'
+                          '  failed = false\n',
+        cases=[
+            Case(assertion=ContextAssertion(key='failed', expected=True))
```
adnrs96: We should test that the error thrown was with message `error` over here

storyruntime/processing/Lexicon.py
adnrs96: I would definitely like to hear about what you think about this

tests/integration/processing/Lexicon.py
TonyRice: On it!

adnrs96:
Had a chat with Tony and things make sense as it is.

tests/integration/processing/Lexicon.py
```diff
@@ -857,8 +857,8 @@
             ),
             Case(
                 append=f'a = {math.pi / 4}\n'
-                       'b = a.tan()\n'
-                       'c = b.approxEqual(value: 1)',
+                'b = a.tan()\n'
+                'c = b.approxEqual(value: 1)',
```
wilzbach: UNRELATED CHANGE!

anukul: 😭😭😭

TonyRice: This change keeps autoformatting. I'll undo that.

Issue #463, opened 2019-08-16
Add support for List.slice


wilzbach:
https://github.com/storyscript/hub-sdk-python/pull/25

judepereira:
@TonyRice will you be doing this this sprint? If yes, please add it to the current sprint. If not, please unassign yourself (so that somebody else can take over when free).

PR #594, opened 2019-10-16
Adding support for List.slice

TonyRice:
https://github.com/storyscript/runtime/issues/463


tests/integration/processing/Mutations.py
```diff
@@ -26,6 +26,14 @@
...
                    assertion=[ContextAssertion(key='a', expected=[1, 2]),
                               ContextAssertion(key='c', expected=[1, 2, 3])]),
+              Case(append='c = a.slice(end: 1)',
+                   assertion=[ContextAssertion(key='c', expected=[1])]),
+              Case(append='c = a.slice(start: 0 end: 1)',
+                   assertion=[ContextAssertion(key='c', expected=[1])]),
+              Case(append='c = a.slice(start: 0 end: 2)',
+                   assertion=[ContextAssertion(key='c', expected=[1, 2])]),
+              Case(append='c = a.slice(start: 1 end: 2)',
+                   assertion=[ContextAssertion(key='c', expected=[2])]),
```
wilzbach: Maybe add a test that ensures that the original data object wasn't modified?

judepereira: Sorry, but using index 2 above should throw a Storyscript runtime error. Index 2 is not available.

wilzbach: Really? Why can't we be forgiving like in other languages?

PR #595, opened 2019-10-16
Update to HttpUtils.add_params_to_url 

TonyRice:
This updates HttpUtils.add_params_to_url to encode values that are an array, as a dictionary with multiple values.

This will fix an issue where a service argument specifies a list argument
with the 'in: query' option. With this update, the action passed with a list
argument, will automatically convert the parameters properly resulting
in a url that actually reflects the http spec

myservice action something: ['hello', 'world']

http://myservice:5000/path?something=hello&something=world


storyruntime/utils/HttpUtils.py
```diff
@@ -53,4 +54,18 @@ def add_params_to_url(url, params: dict):
...
+                for lv in value:
+                    parts.append(
+                        f'{key}={urllib.parse.quote(str(lv))}'
+                    )
+            else:
+                parts.append(
+                    f'{key}={urllib.parse.quote(str(value))}'
+                )
+
+        return f'{url}?{("&".join(parts))}'
```
wilzbach: Let's use urlencode instead here please:

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode

> Convert a mapping object or a sequence of two-element tuples, which may contain str or bytes objects, to a percent-encoded ASCII text string

> The resulting string is a series of key=value pairs separated by '&' characters, where both key and value are quoted using the quote_via function. By default, quote_plus() is used to quote the values, which means spaces are quoted as a '+' character and ‘/’ characters are encoded as %2F, which follows the standard for GET request

```suggestion
        return url + urllib parse.urlencode(parts)
```

storyruntime/utils/HttpUtils.py
```diff
@@ -53,4 +54,18 @@ def add_params_to_url(url, params: dict):
...
+            key = urllib.parse.quote(key)
+            if type(value) is list:
+                for lv in value:
+                    parts.append(
+                        f'{key}={urllib.parse.quote(str(lv))}'
+                    )
+            else:
+                parts.append(
+                    f'{key}={urllib.parse.quote(str(value))}'
+                )
```
wilzbach:
```suggestion
                parts.append((key, str(value)))
```

storyruntime/utils/HttpUtils.py
```diff
@@ -53,4 +54,18 @@ def add_params_to_url(url, params: dict):
...
             return url
 
-        return f'{url}?{urlencode(params)}'
+        parts = list()
+
+        for key, value in params.items():
+            key = urllib.parse.quote(key)
+            if type(value) is list:
+                for lv in value:
+                    parts.append(
```
wilzbach:
```suggestion
                    parts.append((key, str(lv)))
```

tests/unit/utils/HttpUtils.py
```diff
@@ -67,3 +67,6 @@ def test_add_params_to_url():
     assert HttpUtils.add_params_to_url('asyncy.com', {
         'a': 1, 'b': 'c'
     }) == 'asyncy.com?a=1&b=c'
+    assert HttpUtils.add_params_to_url('asyncy.com', {
+        'a': [1, 2, 3], 'b': 'c'
+    }) == 'asyncy.com?a=1&a=2&a=3&b=c'
```
wilzbach: should test that quoting actually works

storyruntime/utils/HttpUtils.py
```diff
@@ -53,4 +54,14 @@ def add_params_to_url(url, params: dict):
         if len(params) == 0:
             return url
 
-        return f'{url}?{urlencode(params)}'
+        parts = list()
```
wilzbach: More Pythonic:

```suggestion
        parts = []
```

wilzbach:
> myservice action something: ['hello', 'world']
> http://myservice:5000/path?something=hello&something=world

It would be nice to have such high-level integration tests :/

PR #596, opened 2019-10-16
Add ability for file service to retrieve details about file

TonyRice:
This currently allows you to determine when a file was created, accessed, modified, and it's file size.


storyruntime/processing/internal/File.py
```diff
@@ -100,6 +100,38 @@ def clean_path(story, path):
...
+                message=f'Failed to retrieve file details: '
+                        f'No such file or directory: '
+                        f'\'{clean_path(story, path)}\'',
+                        story=story, line=line
+            )
+        stats = os.stat(path)
+        return {
+            'accessed': stats.st_atime,
+            'modified': stats.st_mtime,
+            'created': stats.st_ctime,
```
wilzbach: No.
> On some systems (like Unix) is the time of the last metadata change

storyruntime/processing/internal/File.py
```diff
@@ -100,6 +100,38 @@ def clean_path(story, path):
...
+        }
+    except FileNotFoundError:
+        raise StoryscriptError(
+            message=f'Failed to read file: No such file: '
+            f'\'{clean_path(story, path)}\'',
+            story=story, line=line
+        )
+    except IOError as e:
+        raise StoryscriptError(message=f'Failed to read file: {e}',
+                               story=story, line=line)
```
wilzbach: image

judepereira:
@TonyRice What was your motivation behind adding this functionality?

TonyRice:
> @TonyRice What was your motivation behind adding this functionality?

This was actually a suggestion @wilzbach brought up during my last round of http service enhancements.

storyruntime/processing/internal/File.py
```diff
@@ -100,6 +100,48 @@ def clean_path(story, path):
...
+        }
+
+        if stats.st_atime and stats.st_atime > 0:
+            details['accessed'] = stats.st_atime
+
+        if stats.st_mtime and stats.st_mtime > 0:
+            details['modified'] = stats.st_mtime
+
+        if stats.st_ctime and stats.st_ctime > 0:
+            details['created'] = stats.st_ctime
```
wilzbach: This SHOULD NOT be called created.

> On some systems (like Unix) is the time of the last metadata change

https://docs.python.org/3/library/stat.html#stat.ST_CTIME

storyruntime/processing/internal/File.py
wilzbach: How was this addressed?

storyruntime/processing/internal/File.py
```diff
@@ -126,7 +127,12 @@ def clean_path(story, path):
         if stats.st_mtime and stats.st_mtime > 0:
             details['modified'] = stats.st_mtime
 
-        if stats.st_ctime and stats.st_ctime > 0:
+        # Note: We currently only support created
+        # when running on Windows. This is because
+        # of os limitations. If this ever changes
+        # we can update accordingly
+        if stats.st_ctime and stats.st_ctime > 0 \
+                and platform.system() == 'Windows':
```
wilzbach: We don't even support Windows. The runtime is only supposed to run on Linux.

TonyRice:
I will remove it. 
null
Sent via Migadu.com, world's easiest email hosting

PR #598, opened 2019-10-16
Updating TypeUtils.safe_type

TonyRice:
This updates safe_type to ensure a dictionary doesn't have any CaseInsensitiveDicts within its. This is currently causing issues for type casting.


storyruntime/utils/TypeUtils.py
```diff
@@ -49,8 +49,21 @@ def safe_type(o):
         """
         if o is None:
             return None
-        elif isinstance(o, CaseInsensitiveDict):
+
+        if isinstance(o, CaseInsensitiveDict):
             return dict(o.items())
```
wilzbach: Shouldn't we call the safe function on its items here like on dict below?

tests/unit/utils/TypeUtils.py
```diff
@@ -67,3 +69,27 @@ def test_safe_type_internal_command(magic):
...
+    }
+
+    assert _dict == TypeUtils.safe_type(
+        CaseInsensitiveDict(_dict)
+    )
+    assert _dict2 == TypeUtils.safe_type({
+        'hello': 'world',
+        'WORLD': 'hello',
+        'else': CaseInsensitiveDict(_dict)
+    })
```
wilzbach: Those are two different tests and hence should be in different test methods.

storyruntime/utils/TypeUtils.py
```diff
@@ -49,8 +49,21 @@ def safe_type(o):
...
+        elif isinstance(o, dict):
+            for key, val in o.items():
+                o[key] = TypeUtils.safe_type(val)
+
+            return o
+        elif isinstance(o, list):
+            new = list()
+            for val in o:
+                new.append(
+                    TypeUtils.safe_type(val)
```
wilzbach: https://stackoverflow.com/questions/3487802/which-is-generally-faster-a-yield-or-an-append

adnrs96: Just wanted to mention that using a list comprehension (`[TypeUtils.safe_type(d) for d in o]`) over here instead of a generator would have been even more faster.

Issue #271, opened 2019-04-17
Expression on more types


wilzbach:
The following operations are currently all valid in Storyscript:

Operations on Boolean:

```
a = true + true 
a = 1 + true 
a = 1 - true 
```

Operations on Time:

```
a = 5s + 10m
b = 10m - 5s
b = 10m / 5s
b = 10m * 5s
b = 10m % 5s
```

Operations on Lists:

```
a = [1, 2, 3] + [4, 5]
```

Operations on Maps:

```
a = {"a": 1} } + {"b": 2}
```

I can also restrict the set of allowed operands, but the engine and compiler should match up.

wilzbach:
Documentation can be found here: https://github.com/asyncy/docs.asyncy.com/pull/55
SS behavior is specified here in __a lot__ of tests:

https://github.com/storyscript/storyscript/tree/master/tests/integration/compiler/semantics/types

PR #599, opened 2019-10-16
sum expression on Map

TonyRice:
See https://github.com/storyscript/runtime/issues/271


tests/integration/processing/Lexicon.py
```diff
@@ -1423,6 +1423,25 @@
             )
         ]
     ),
+    Suite(
+        preparation_lines='a = [1, 2, 3]\n'
+                          'b = a + [3, 4, 5]',
```
anukul: `list + list` is already covered in the test above this, I guess these should be changed to `dict + dict`, since this PR adds that.

storyruntime/utils/Resolver.py
```diff
@@ -193,6 +193,9 @@ def expression(self, item):
                 if type(r) in (int, float, list) and \
                         type(result) in (int, float, list):
                     result += r
+                elif type(r) is dict and type(result) is dict:
+                    result = dict(result)
```
anukul: Since `type(result) is dict`, we can probably remove this cast

wilzbach: @anukul this creates a shallow copy, s.t. the original object isn't updated

tests/integration/processing/Lexicon.py
```diff
@@ -1423,6 +1423,25 @@
...
+        cases=[
+            Case(
+                assertion=ContextAssertion(key='b',
+                                           expected=[1, 2, 3, 3, 4, 5])
+            )
```
wilzbach: Can we add an assertion that `a` stayed unmodified?

tests/integration/processing/Lexicon.py
```diff
@@ -1423,6 +1423,25 @@
...
+    ),
+    Suite(
+        preparation_lines='a = [1, 2, 3] + [3, 4, 5]',
+        cases=[
+            Case(
+                assertion=ContextAssertion(key='a',
+                                           expected=[1, 2, 3, 3, 4, 5])
+            )
+        ]
+    ),
```
wilzbach: Tests are for lists, but the PR introduces `+` for Maps.

PR #600, opened 2019-10-16
Minor file service error additions

TonyRice:


anukul:
Can we add tests for the `file` microservice?
I see a bunch of them mixed up in `test_mutation`, maybe this is is a good chance to move them to a separate `test_file` suite, and also increase the coverage for this PR.

PR #602, opened 2019-10-17
Adding StringUtils.bytes_to_bytestr()

TonyRice:
See: https://github.com/storyscript/runtime/issues/550


storyruntime/utils/StringUtils.py
```diff
@@ -16,3 +16,11 @@ def truncate(result, max_bytes: int):
                      f'({truncated_len} bytes truncated)'
 
         return result
+
+    @staticmethod
+    def bytes_to_bytestr(data: bytes):
+        bytestr = ''
+        for byte in data:
+            bytestr += chr(byte)
```
wilzbach: https://stackoverflow.com/questions/30245397/why-is-a-list-comprehension-so-much-faster-than-appending-to-a-list

judepereira:
@TonyRice Sorry, but my original issue was more of a question (see the tags on #550). We haven't arrived at a solution for this yet.

TonyRice:
> @TonyRice Sorry, but my original issue was more of a question (see the tags on #550). We haven't arrived at a solution for this yet.

I should have clarified, this is more of a proposal for this problem. This solution more or less converts the bytes to a "bytestr" which is what happens when you do something like print(raw_binary) within the python CLI. More or less a human-readable string. For example:

```
mystr = '\xc69\xb6JV\x89Z\x92\xa6\xb3'
bytes = b'\xc69\xb6JV\x89Z\x92\xa6\xb3'
```

judepereira:
Hmm, I don't quite like this proposal of bytestr. Let's continue this on the issue thread?

TonyRice:
Definitely! I'll go ahead and move things over there

PR #608, opened 2019-10-17
Add list.join mutation

### LINKED ISSUE #588: Implement List.join
https://github.com/storyscript/hub-sdk-python/pull/55

Should stringify each element and then concatenate them with `sep`.

anukul:
Closes #588

### DIFF
--- a/storyruntime/processing/mutations/ListMutations.py
+++ b/storyruntime/processing/mutations/ListMutations.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 import random
 
+from storyruntime.utils.TypeResolver import TypeResolver
+
 
 class ListMutations:
 
@@ -92,3 +94,14 @@ class ListMutations:
                 value[i] = by
 
         return value
+
+    @classmethod
+    def join(cls, mutation, value, story, line, operator):
+        sep = story.argument_by_name(mutation, 'sep')
+        return sep.join([
+            TypeResolver.type_cast(item=v, type_={
+                '$OBJECT': 'type',
+                'type': 'string'
+            })
+            for v in value
+        ])
--- a/tests/integration/processing/Mutations.py
+++ b/tests/integration/processing/Mutations.py
@@ -558,6 +558,22 @@ async def test_string_mutations(suite: Suite, logger, run_suite):
                     expected=[1, 42, 42, 3, 4, 4, 5, 5]
                 )
             ),
+            Case(
+                append='s = arr.join(sep: "x")',
+                assertion=ContextAssertion(key='s', expected='1x2x2x3x4x4x5x5')
+            )
+        ]
+    ),
+    Suite(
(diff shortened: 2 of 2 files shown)


anukul:
`str(value)` or `json.dumps(value)`?

tests/integration/processing/Lexicon.py
```diff
@@ -971,6 +971,9 @@
             Case(append='arr = arr.replace(item: 2 by: 42)',
                  assertion=ContextAssertion(
                      key='arr', expected=[1, 42, 42, 3, 4, 4, 5, 5])),
+            Case(append='s = arr.join(sep: "x")',
+                 assertion=ContextAssertion(
+                     key='s', expected='1x2x2x3x4x4x5x5')),
         ]),
```
wilzbach: Let's add a test for non string values like e.g. array or ints.

wilzbach:
> str(value) or json.dumps(value)?

`str`.

tests/integration/processing/Lexicon.py
anukul: This is a test for ints

anukul:
Okay this is a little weird..
- `str({"a": True}) = "{'a': True}"`

I guess the "right" way to solve it would introducing our own types that define their string representation explicitly, but `json.dumps` solves it for now

cc @judepereira

tests/integration/processing/Lexicon.py
wilzbach: Ah I wanted to say: "array of ints"

wilzbach:
> I guess the "right" way to solve it would involve introducing our own types that define their string representation explicitly,

Yup this is crucial see e.g. -> https://github.com/storyscript/runtime/issues/327

storyruntime/processing/mutations/ListMutations.py
```diff
@@ -92,3 +93,8 @@ def replace(cls, mutation, value, story, line, operator):
                 value[i] = by
 
         return value
+
+    @classmethod
+    def join(cls, mutation, value, story, line, operator):
+        sep = story.argument_by_name(mutation, 'sep')
+        return sep.join([json.dumps(v) for v in value])
```
judepereira: From what we discussed - can you simulate a cast to a string? This will make sure that all the safe_type code written executes too, and that internal implementation details don't leak accidentally (since we already have code for this).

anukul: Updated, thanks for reminding

anukul:
image

Sorry about the huge diff, it's largely a result of segregating tests (yet again) and adopting consistent indentation (either single line or a singular, specific, multi-line format)

judepereira:
@anukul can you please split this PR into another one with just the list.join? I'd love to see the tests you're adding in Lexicon for this, but can't because far too much has changed.

anukul:
@judepereira updated. this is now based on top of https://github.com/storyscript/runtime/pull/622, and includes a single commit which should be easy to review

anukul:
Fixed conflicts in #622, rebased this again

PR #628, opened 2019-10-24
Lock dependencies

### LINKED ISSUE #555: Dependencies should not be duplicated between tox.ini and setup.py
Currently we maintain the pinning of test dependencies in both the `tox.ini` and the `setup.py`. These got out of sync and tests were passing in CI via `tox`, but I could not manually run tests locally using `setuptools`.

https://github.com/storyscript/runtime/pull/554 has fixed the version discrepancy but manually maintaining these two lists will continue to cause us problems.

tox.ini: https://github.com/storyscript/runtime/blob/master/tox.ini#L8-L11
setup.py: https://github.com/storyscript/runtime/blob/master/setup.py#L25-L29

### LINKED ISSUE #604: Lock dependencies with a lock file
Storyscript no longer pins its dependencies (see e.g. https://github.com/storyscript/storyscript/releases/tag/0.25.1).

However, the runtime should make sure that all versions used during the run of the test suite match up with what ends up being deployed.

This can be done with e.g. pipenv, but there are many approaches.

anukul:
Closes #604, closes #555 

New workflow:

- Development
1. Update dependencies in `setup.py`
2. Run `pip-compile` to update `requirements.txt`
3. Run `pip-sync` to sync virtual environment

- Deployment
1. Run `pip-sync` to install dependencies (or `pip install -r requirements.txt`?)


setup.py
```diff
@@ -15,35 +15,42 @@
 readme = io.open('README.md', 'r', encoding='utf-8').read()
 
 setup(
-    name='storyscript-platform-engine',
-    description='The engine of the Storyscript platform',
+    name='storyruntime',
```
wilzbach:
```suggestion
    name='story-runtime',
```

https://pypi.org/project/story-runtime

.circleci/config.yml
```diff
@@ -20,40 +20,19 @@ jobs:
...
-            python3 -m venv venv
-            . venv/bin/activate
-            pip3 install tox
-            env
-            python3 setup.py install
-
-      - save_cache:
-          paths:
-            - ./venv
-          key: v1-deps-{{ checksum "setup.py" }}
```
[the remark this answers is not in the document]
anukul: I think installing dependencies into this environment is unnecessary, since the actual testing happens in isolation in tox environments, and they are not cached at the moment.
If you take a look at circleci, the build times remain unaffected.

tox.ini
```diff
@@ -3,24 +3,18 @@ envlist = py37,pep8
 
 
 [testenv:py37]
```
jayvdb: remove the `:py37` here.  It is useless, and makes it more difficult to use tox effectively

We should also have `py38` in envlist, to start testing with Python 3.8, but that is not particularly important as people can run `tox -e py38` on Python 3.8 and it will run correctly, if this `:py37` is removed.

tox.ini
```diff
@@ -3,24 +3,18 @@ envlist = py37,pep8
...
+# recreate because tox does not parse requirements to detect changes
+recreate = true
+deps = -rrequirements.txt
+extras = py37
```
jayvdb: call this `test`, and update accordingly in `setup.py`.  An extra named `py37` for tests is confusing, as the items in the extra have nothing to do with Python 3.7.

tox.ini
```diff
@@ -3,24 +3,18 @@ envlist = py37,pep8
...
     coverage xml
 
 
 [testenv:pep8]
-deps =
-    flake8==3.5.0
-    flake8-quotes==1.0.0
-    flake8-import-order==0.18
-    pep8-naming==0.7.0
+extras = pep8
```
jayvdb: not as bad as `py37`, but maybe call it `style` or `stylecheck` to avoid locking the name into a specific style guide or component name (which we probably wont comply with https://github.com/storyscript/community/issues/17), and we are possibly going to be moving to using `black` as the primary style tool (https://github.com/storyscript/community/issues/16).

.circleci/config.yml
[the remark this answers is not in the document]
jayvdb: pip caching is only a matter of ensuring `$PIP_CACHE_DIR` is set and consistent in all layers of virtualenvs, or setting `$XDG_CACHE_HOME` from which pip will then derive its cache dir.  But IMO caching can be tidied up / set up, in a subsequent PR after the locking is in place.  Its a performance thing, not a correctness and stability thing.

It definitely doesnt make sense to be installing all deps outside of tox and then also inside of tox - it is a time consuming process, even if the packages are all cached.

The one annoyance is that usually `codecov` is run outside of `tox`, which means some of the deps (`coverage` & `codecov`) are needed outside.  However than can be avoided by using tox `commands_post` like at https://github.com/coala/coala-bears/blob/master/tox.ini#L112 (the `--flags=..` there depends on my plugin `tox-backticks`)

setup.py
```diff
@@ -15,35 +15,42 @@
...
         'requests==2.21.0'  # Used for structures like CaseInsensitiveDict.
     ],
+    extras_require={
+        'pep8': [
+            'flake8==3.5.0',
+            'flake8-quotes==1.0.0',
+            'flake8-import-order==0.18',
+            'pep8-naming==0.7.0'
+        ],
+        'py37': [
```
jayvdb: c.f. other comment about this extra name

.circleci/config.yml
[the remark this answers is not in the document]
anukul: Cool, I'll look at the caching, thanks.

The way we have Codecov set up, I think it just collects the generated coverage report later and doesn't have any python dependencies.

tox.ini
```diff
@@ -1,20 +1,21 @@
...
 
 
-[testenv:py37]
+[testenv]
 # recreate because tox does not parse requirements to detect changes
 recreate = true
-deps = -rrequirements.txt
-extras = py37
+deps =
+    !stylecheck: -rrequirements.txt
```
jayvdb: nice workaround.

judepereira:
Nice work @anukul!

PR #637, opened 2019-10-25
Refactor configuration and adopt black

### LINKED ISSUE #625: Adopt consistent formatting via black
https://github.com/psf/black

To avoid personal reformatting which leads to a messed up git history and unnecessary conflicts in other PRs.

1) Do once a reformat with a different author, s.t. it can be filtered from the history
2) Add a `black --check` to the CI
3) Recommend editor setup (e.g. for Vim: `autocmd BufWritePre *.py execute ':Black'`)

There are tons of guides, e.g. https://www.mattlayman.com/blog/2018/python-code-black/

anukul:
Closes #625


anukul:
Should we add a pre-commit hook that runs `black`?

setup.py
```diff
@@ -39,10 +39,7 @@
     ],
     extras_require={
         'stylecheck': [
-            'flake8==3.5.0',
-            'flake8-quotes==1.0.0',
-            'flake8-import-order==0.18',
-            'pep8-naming==0.7.0'
+            'black==19.3b0'
```
wilzbach: Why the beta release?

anukul: https://github.com/psf/black#note-this-is-a-beta-product

Dockerfile
```diff
@@ -1,18 +1,13 @@
...
 
 COPY          . /app/
-WORKDIR       /app
-RUN           chmod +x entrypoint.sh
 RUN           python setup.py install
-ENV           ASSET_DIR /asyncy
-ENV           logger_level info
 
-ENTRYPOINT    ["/app/entrypoint.sh"]
+ENTRYPOINT    ["storyscript-server", "start"]
```
wilzbach: We should probably start testing the docker image to automatically ensure this didn't break anything.
See e.g. https://github.com/storyscript/runtime/issues/639

wilzbach:
> Should we add a pre-commit hook that runs black?

Yep, though IIRC everyone needs to setup this individually in their respective git repository. You can only provide instructions and make the setup easier. It's still worth it to do this, because otherwise no one will try to set this up ;-)

PR #648, opened 2019-10-29
Add string to time type cast

### LINKED ISSUE #261: Implement the time type
Introduced in Storyscript 0.15.0

anukul:
Closes #261 

WIP: Check for `time` type in service call arguments


anukul:
WIP: Check for `time` type in service call arguments

storyruntime/utils/TypeResolver.py
```diff
@@ -109,6 +136,19 @@ def check_type_cast(cls, type_exp, item):
             return int(item)
         elif isinstance(type_exp, types.FloatType):
             return float(item)
+        elif isinstance(type_exp, types.TimeType):
+            cls.assert_type([str], item)
```
wilzbach: Must allow floats as well:

```
a = 1s to time
```

storyruntime/utils/TypeResolver.py
```diff
@@ -109,6 +136,19 @@ def check_type_cast(cls, type_exp, item):
...
+            m = TIME_REGEX.fullmatch(item)
+            if m is None:
+                raise TypeAssertionError()
```
wilzbach: More information should be provided to the user here.

storyruntime/utils/TypeResolver.py
anukul: The `a = 1s to time` example isn't exactly clear to me, since it's already a `time` value.

I assumed we don't allow floats from this
image

anukul: `a = "1s" to time` works though, right?

wilzbach: Ah, I'm sorry for confusing you. I didn't understand myself. I wanted to say "must allow integers".
Here's a better example:

```
a = 1s to any
b = a to time
```

wilzbach:
FYI: I recently proposed an OMS `date` and `duration` type.
https://github.com/microservices/openmicroservices.org/pull/171

Note the more complex representation for `duration` as a single integer isn't enough.

Issue #573, opened 2019-10-11
Fix service metrics error


anukul:
```
Recording resource usage metrics failed\nTraceback (most recent call last):
  File \"/usr/local/lib/python3.7/site-packages/storyscript_platform_engine-0.2.0-py3.7.egg/storyruntime/ServiceUsage.py\", line 130, in start_metrics_recorder
    for service_tag_uuid in current_batch
  File \"/usr/local/lib/python3.7/site-packages/storyscript_platform_engine-0.2.0-py3.7.egg/storyruntime/ServiceUsage.py\", line 88, in get_metrics
    cls.cpu_units(pod['containers'][0]['usage']['cpu'])
IndexError: list index out of range
```

https://github.com/storyscript/runtime/blob/2b31b54f28daebba765de149b0d3140e5b82a93e/storyruntime/ServiceUsage.py#L63-L98

PR #675, opened 2019-10-30
Raise exception for service metrics error

anukul:
Related to https://github.com/storyscript/runtime/issues/573

I can't seem to figure out why any pod would have no containers, so this will help us see that pod's name when such a situation occurs


wilzbach:
@anukul: a story with just internal services (e.g. hello world) has no containers as there's only the global runtime.

PR #699, opened 2019-10-31
Don't try to deploy deleted apps

### LINKED ISSUE #671: Database.get_release_for_deployment: 'NoneType' object is not subscriptable
Sentry Issue: [ENGINE-5Q](https://sentry.io/organizations/asyncy/issues/1296446911/?referrer=github_integration)

```
TypeError: 'NoneType' object is not subscriptable
  File "storyruntime/Apps.py", line 325, in reload_app
    release = await Database.get_release_for_deployment(config, app_id)
  File "storyruntime/db/Database.py", line 130, in get_release_for_deployment
    app_uuid=data["app_uuid"],
```

anukul:
Fixes #671 

Reproduced as:

- Run `story apps delete`. This sets `app.deleted = True` and sends a notification to the runtime which sets the latest release's state to `NO_DEPLOY`.
- When the runtime restarts, it still recieves the app's UUID from `Database.get_all_app_uuids_for_deployment` since that doesn't check for `app.deleted` or `NO_DEPLOY`.
- This UUID is passed on to `Database.get_release_for_deployment` which tries to find the release to deploy for the app. This function discards releases that have `state` set to `NO_DEPLOY`, so it's unable to find an appropriate release, causing the error.

The following weird behavior still remains though:

- If you delete an app that has 10 releases, the delete action will only set the latest release to `NO_DEPLOY`. The runtime will attempt to deploy each of the 9 releases every time it restarts, and set their state to `NO_DEPLOY` one at a time.

Should we make it so that the moment any one release's state is set to `NO_DEPLOY` as a result of the app being deleted, all other releases of the app are also set to `NO_DEPLOY`?

### DIFF
--- a/storyruntime/Apps.py
+++ b/storyruntime/Apps.py
@@ -82,21 +82,6 @@ class Apps:
             )
             return
 
-        if release.deleted:
-            await Database.update_release_state(
-                logger, config, app_id, release.version, ReleaseState.NO_DEPLOY
-            )
-            logger.warn(
-                f"Deployment halted {app_id}@{release.version}; "
-                f"deleted={release.deleted}; "
-                f"maintenance={release.maintenance}"
-            )
-            logger.warn(
-                f"State changed to NO_DEPLOY for {app_id}@"
-                f"{release.version}"
-            )
-            return
-
         await Database.update_release_state(
             logger, config, app_id, release.version, ReleaseState.DEPLOYING
         )
@@ -334,6 +319,10 @@ class Apps:
                     f"expected {config.APP_ENVIRONMENT}, "
                     f"but got {release.app_environment})"
                 )
+                return
+
+            if release.deleted:
+                glogger.warn(f"Not deploying app {app_id} since it is deleted")
                 return
 
             if release.state == ReleaseState.FAILED.value:
--- a/storyruntime/db/Database.py
+++ b/storyruntime/db/Database.py
@@ -51,6 +51,7 @@ class Database:
             from releases
                      inner join apps on releases.app_uuid = apps.uuid
             where environment = $1
+            and not apps.deleted
             group by app_uuid;
             """
             return await con.fetch(query, config.APP_ENVIRONMENT.value)
--- a/tests/unit/Apps.py
+++ b/tests/unit/Apps.py
@@ -220,7 +220,7 @@ async def test_reload_app_no_story(patch, config, logger, db, async_mock):
         always_pull_images=False,
         app_dns="app_dns",
         state="QUEUED",
-        deleted=True,
+        deleted=False,
         owner_uuid="owner_uuid",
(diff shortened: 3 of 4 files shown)


judepereira:
@anukul Since we're already running an inner join with apps, why not check for maintenance and deleted? No need to play with the state.

judepereira:
That will also solve your weird behaviour.

anukul:
Sure, in that case should we just remove the `NO_DEPLOY` state?

judepereira:
In your SQL update for this PR? Yes.

anukul:
Beyond demarcating the latest release of deleted apps, what is its significance?

If we never call `get_release_for_deployment` on a deleted app, I'm not sure where this state will be used.

judepereira:
`NO_DEPLOY` may also be set on a release by the CLI, where the user does not want to deploy, but rather just make config changes successively.

anukul:
At the moment it looks like the CLI just sends an empty payload, and the database [sets](https://github.com/storyscript/database/blob/e6f374baad42bda600b077f4b4f2fef0d2ef1503/deploy/bootstrap.sql#L490-L493) it to the previous release's payload.

judepereira:
> At the moment it looks like the CLI just sends an empty payload, and the database sets it to the previous release's payload.

Yes, but this causes a deployment in the runtime.
There is a use case for not deploying a release. However, I don't think it's implemented in the CLI yet.

anukul:
Cool, in that case I've removed the `deleted` key from the `Release` named tuple, since it was just tied to `NO_DEPLOY` which we will no longer use for determining deployment of deleted apps after this PR is merged.

judepereira:
@anukul what happens when an app is deleted while the runtime is running? I believe that the deleted check is still valuable and needs to continue residing in Release.

anukul:
Right, I forgot to account for that. :(

Currently, it will first terminate the release (`TERMINATED`), then set it to `NO_DEPLOY`.
But since that's not relevant anymore, I think we should just stop after destroying if the app is deleted.

anukul:
I can add `Release.deleted` back in, but I don't see where it will be used.

anukul:
@judepereira updated

storyruntime/Apps.py
```diff
@@ -82,21 +82,6 @@ def get_app_config(cls, raw):
             )
             return
 
-        if release.deleted:
```
judepereira: Why'd you remove this?

anukul: It's checked on line 324, and the state change is not needed

judepereira: Makes sense, since deploy_release is called from two places:
1. At boot time - you addressed the deleted by updating the query in get_all_apps....
2. At runtime, when a notification is received - you addressed this reload_app.

PR #700, opened 2019-11-01
Don't insert None for missing optional arguments

anukul:
Fixes https://github.com/storyscript/runtime/issues/693

This PR primarily fixes the bug that is currently breaking redis.
Unit test for `execute_http` is nearly incomprehensible to me wrt making changes for this, so diff coverage is low. In an ideal scenario, IMO this would undergo heavy refactoring.


storyruntime/Story.py
```diff
@@ -269,19 +269,19 @@ def function_line_by_name(self, function_name):
         line_number = self.app.stories[self.name]["functions"][function_name]
         return self.line(line_number)
 
-    def argument_by_name(self, line, argument_name, encode=False):
-        args = line.get("args", line.get("arguments", line.get("arg")))
+    def argument_by_name(
+        self, line, argument_name, encode=False, throw_if_not_found=False
+    ):
+        args = line.get("args")
         if args is None:
```
anukul: When does this case occur?

storyruntime/processing/Services.py
```diff
@@ -496,8 +496,13 @@ def smart_insert(
         request_body_fields_count = 0
 
         for arg in args:
-            value = story.argument_by_name(line, arg)
             arg_conf = args[arg]
+            try:
+                value = story.argument_by_name(
+                    line, arg, throw_if_not_found=True
+                )
+            except ArgumentNotFound:
```
anukul: Assuming that we never recieve a payload where a required argument is missing, because the compiler would raise an error in such a situation.

storyruntime/Story.py
judepereira: Not sure. I think this was related to some compiler change, and exists for backwards compatibility.

storyruntime/processing/Lexicon.py
```diff
@@ -54,7 +54,7 @@ class Lexicon:
                 assign={"paths": line.get("output")},
             )
 
-            return Lexicon.line_number_or_none(story.line(line.get("enter")))
+            return line.get("enter")
```
judepereira: Can we separate this change out into another PR?

storyruntime/processing/Lexicon.py
```diff
@@ -70,7 +70,7 @@ class Lexicon:
                     line["ln"], output=output, assign=line.get("output")
                 )
 
-            return Lexicon.line_number_or_none(story.line(line.get("next")))
+            return line.get("next")
```
judepereira: Same. See above.

storyruntime/processing/Lexicon.py
```diff
@@ -555,7 +555,7 @@ def throw(logger, story, line):
         1. From when blocks - no value may be returned
         2. From function blocks - one value may be returned
         """
-        args = line.get("args", line.get("arguments"))
+        args = line.get("args")
```
judepereira: Exists for backwards compatibility. Let it stay there, unless you have a strong reason to remove it.

storyruntime/processing/Services.py
```diff
@@ -405,10 +409,6 @@ def raise_for_type_mismatch(cls, story, line, name, value, arg_conf):
         """
         t = arg_conf.get("type", "any")
 
-        if value is None and not arg_conf.get("required", False):
```
judepereira: Umm? If the value is none, then the last line of this method will throw an exception.

tests/conftest.py
```diff
@@ -85,9 +85,9 @@ def echo_line():
         LineConstants.command: "echo",
         "args": [
             {
-                "$OBJECT": "argument",
+                "$OBJECT": "arg",
```
judepereira: This could have been done separately.

tests/unit/processing/Lexicon.py
```diff
@@ -45,7 +45,6 @@ def line():
 
     output = MagicMock()
     patch.object(Services, "execute", new=async_mock(return_value=output))
-    patch.object(Lexicon, "line_number_or_none")
```
judepereira: Mixing up two features in this PR.

storyruntime/processing/Services.py
anukul: It should only throw an exception if value is `null` and type is not `any`.

storyruntime/Story.py
wilzbach: Yes, this was changed quite a while ago.

storyruntime/processing/Services.py
anukul: unless the spec allows other cases

judepereira: So shouldn't we keep this check? If the value is none, and this is not a required argument, just return.

storyruntime/processing/Lexicon.py
wilzbach: This has been changed a long time ago and since this week only code compiled with 0.25 can run on the runtime, so I don't see a reason to keep it?

storyruntime/processing/Services.py
wilzbach: https://docs.storyscript.io/storyscript/advanced/#null-value

> using null as an argument to a service/function/built-in will result in a runtime error if the argument type is not any

anukul: @judepereira it will return on line 453

wilzbach:
BTW with `variadic` unknown arguments are allowed - https://github.com/storyscript/runtime/issues/710.

anukul:
Cool, I'll add that in the next PR.

Just rebased this and removed the other commits, deployed to staging and confirmed working on the following story:

```
redis set key:"count" value:0
when http server listen path:"/counter" as request
    count = redis increment key:"count"
    request write content:"{count}"
```

Notice the missing `by` argument in `redis set`. This is currently failing on production.

PR #714, opened 2019-11-06
TypeUtils Fix

TonyRice:
There was a change introduced a little while ago that broke conversion for CaseInsensitiveDicts, this fixes that.

See https://github.com/storyscript/runtime/issues/715


TonyRice:
> Needs tests

Adding tests

PR #721, opened 2019-11-07
[do not merge] Directly use `services` list in App.start_services.

### LINKED ISSUE #663: Directly use `services` list in App.start_services
There's no need to iterate through all lines anymore. The `services` field provides all required information.

judepereira:
Closes #663.


storyruntime/App.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, app_data: AppData):
         self.environment = CaseInsensitiveDict(data=self.environment)
         self.stories = release.stories["stories"]
         self.entrypoint = release.stories["entrypoint"]
+        self.services_used = release.stories["services"]
```
anukul: Can we use `App.services` here?
Since that also seems to be populated from the same source
https://github.com/storyscript/runtime/blob/61b0afa171ec14484c8ee8d8b11ab235b2c5cdcd/storyruntime/Apps.py#L213-L246

judepereira: True. It doesn't really matter though. Any strong argument for changing it?

anukul: Not sure, just cleaner code.
Should I do it?

judepereira: Sure, you can do it.

